PR #19, opened 2016-11-22
SAM FLAGs

alumi:
Add util functions for handling FLAG field of SAM format.

Original FLAG is an integer interpreted as a bitset. 
New functions in `cljam.util.sam-util` convert bitset <=> set of keywords as below.

```clojure
(cljam.util.sam-util/encode-flags 83) => #{:multiple :properly-aligned :reversed :first}
(cljam.util.sam-util/decode-flags #{:multiple :properly-aligned :reversed :first}) => 83
```


totakke:
Good implementation. I think there is no problem with this PR being merged.

PR #22, opened 2017-01-25
Improve performance of reading BAM files.

alumi:
This PR improves performance of reading alignments from a BAM file.

Decoding 4-bit encoded sequences takes approximately **70%** of time during reading BAM.
I modified to use static byte-to-chars table to make decoding about **30x faster**, therefore entire `cljam view` command got about **3x faster**.

I also modified project.clj to update dependencies.

After this modification, the largest bottleneck will be decoding optional fields which is about 4x slower than decoding a sequence.


src/cljam/util/sam_util.clj
```diff
@@ -318,6 +320,26 @@
           (aset-byte result i (byte (bit-or h l)))
           (recur (inc i)))))))
 
+(def ^:const ^:private seq-byte-table
+  (let [nibble-table "=ACMGRSVTWYHKDBN"]
+    (->> (for [i nibble-table j nibble-table] [i j])
+         (apply concat)
+         clojure.string/join)))
```
totakke: `clojure.string/join` -> `cstr/join`

alumi:
Thank you for the comment. I've fixed it and rebased.

totakke:
Thanks. This improvement has just merged 👍

PR #27, opened 2017-02-09
Fix bugs in mpileup.

alumi:
#### Summary

This PR fixes some problems in mpileup command.

#### Fixes

- Fix: may cause StackOverflow when piling up long sequence.
- Fix: quality score may be incorrect for bases with complex cigar.
- Fix: output deletions.
- Memoize parsing cigar string into sequence of indices. Parsing itself also get slightly faster.

#### Known issues

There is a difference between cljam and samtools for adjacent indels.

*input SAM*

```
@SQ SN:ref  LN:20
r001 0 ref 2 60  1M5D5I  * 0 0 ATGCAT  IIIIII
```

*samtools 1.3.1 output*
```
ref	2	N	1	^]A-5NNNNN	I
ref	3	N	1	*	I
ref	4	N	1	*	I
ref	5	N	1	*	I
ref	6	N	1	*	I
ref	7	N	1	*+5GCATC$	I
```

*cljam output*
```
ref	2	N	1	A-5NNNNN	I
ref	3	N	1	*	~
ref	4	N	1	*	~
ref	5	N	1	*	~
ref	6	N	1	*	~
ref	7	N	1	*+5TGCAT	~
```

I'm not certain about the spec, but I guess the last pileup should be `*+5TGCAT$`.
The last `C` at 7th base is an artifact from first nibble of quality `IIIIII` caused by buffer overrun.
Note that we're omitting `^`, mapq and `$` to mark start / end of reads.


totakke:
The pointed INDEL problem has existed since before. It should be noted in cljam issue.

alumi:
Thank you. I opened an issue #28.

PR #29, opened 2017-02-14
Improve performance of reading vcf file.

alumi:
This PR improves performance of reading VCF file.

`cljam.vcf.reader` is slow when reading VCF files like [1000genome](http://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20130502/) which contains enormous number of genotype fields (> 2500).

This PR fixed 3 problems to make it more than **2x faster**.
- Cache vector of keywords. `keyword` calls `intern` for symbol and keyword which causes locking for static state of runtime.
- Use `apply hash-map` instead of `merge zipmap`. I don't know why but `zipmap` uses persistent map. `hash-map` will use transient hash map and this would be faster.
- Replace `clojure.string/split` with `java.lang.String/split`. [Some implementations](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/lang/String.java#2317) optimize `java.lang.String/split` with single character as delimiter  to avoid using `java.util.regex.Pattern`.

### DIFF
--- a/src/cljam/vcf/reader.clj
+++ b/src/cljam/vcf/reader.clj
@@ -3,7 +3,8 @@
   https://samtools.github.io/hts-specs/ for the detail VCF specifications."
   (:require [clojure.string :as cstr]
             [camel-snake-kebab.core :refer [->kebab-case-keyword]]
-            [cljam.util :refer [str->long]]))
+            [cljam.util :refer [str->long]])
+  (:import [clojure.lang LazilyPersistentVector]))
 
 ;; VCFReader
 ;; ---------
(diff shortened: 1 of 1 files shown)


src/cljam/vcf/reader.clj
```diff
@@ -93,28 +94,32 @@
 ;; ------------------
 
 (defn- parse-data-line
-  [line header]
-  (let [[fields gt-fields] (->> (cstr/split line #"\t")
+  [line header kws]
+  (let [[fields gt-fields] (->> (.split ^String line "\t")
+                                clojure.lang.LazilyPersistentVector/createOwning
```
totakke: These two lines have potentially degradation risk. Please add comment describing that this implementation is for performance issue.

alumi: Thank you for the advice.
I've added comments about performance and link to this PR.

totakke:
Perfect! 💯

PR #31, opened 2017-03-03
2bit-encoded reference reader

alumi:
Added a reader for [2bit reference file format](http://genome.ucsc.edu/FAQ/FAQformat.html#format7).
Also added test files converted by [faToTwoBit](https://genome.ucsc.edu/goldenpath/help/blatSpec.html#faToTwoBitUsage)

.2bit is a compact randomly-accessible format of DNA sequences.

```clojure
(with-open [r (cljam.twobit/reader "PATH/TO/2BIT")]
  (cljam.twobit/read-sequence r {:chr "chr1" :start 1 :end 1000}))
=> "ATGC..."
```

Maybe we should have some protocols or some abstractions for range-based querying (`read-sequence`) for FASTA and 2bit.


totakke:
Thanks!

> Maybe we should have some protocols or some abstractions for range-based querying (`read-sequence`) for FASTA and 2bit.

I think so, too. `SeqReader` or `RandomSeqReader` looks good.

Issue #2, opened 2014-04-04
Removing dupricate reads


totakke:
A new feature for removing dupricate reads like Picard MarkDupricates.

abhi18av:
I'm curious about this one after the referenced PR, why is this issue kept open? 🤔

totakke:
Right. Already resolved.

abhi18av:
Cool :)

PR #32, opened 2017-03-06
Add functions for deduplicating paired-end BAM.

alumi:
This PR adds a function `cljam.dedup/dedup` and a transducer `cljam.dedup/dedup-xform` for removing PCR duplications in paired-end BAM files.

```clojure
(sequence
  (dedup/dedup-xform)
  [{:qname "r1" :flag 99 :rname "ref" :pos 1 :mapq 60 :rnext "=" :pnext 10 :tlen 5 :qual "IIIII"}
   {:qname "r2" :flag 99 :rname "ref" :pos 1 :mapq 60 :rnext "=" :pnext 10 :tlen 5 :qual "IIIIH"}
   {:qname "r1" :flag 147 :rname "ref" :pos 10 :mapq 60 :rnext "=" :pnext 1 :tlen -5 :qual "IIIII"}])
 => [{:qname "r1" :flag 99 :rname "ref" :pos 1 :mapq 60 :rnext "=" :pnext 10 :tlen 5 :qual "IIIII"}
     {:qname "r1" :flag 147 :rname "ref" :pos 10 :mapq 60 :rnext "=" :pnext 1 :tlen -5 :qual "IIIII"}]
```

Its algorithm is similar to `rmdup` command of samtools.
As mentioned in #2, `MarkDuplicates` command of Picard implements more complex processes such as 2-pass filtering and file-backed indexing to remove not only PCR duplications but also optical duplications.

### DIFF
--- a/src/cljam/dedupe.clj
+++ b/src/cljam/dedupe.clj
@@ -1,0 +1,70 @@
+(ns cljam.dedupe
+  (:refer-clojure :exclude [dedupe])
+  (:require [com.climate.claypoole :as cp]
+            [cljam.io :as io]
+            [cljam.bam :as bam]
+            [cljam.util.sam-util :as sam-util]))
+
+(defn- refs->regions [refs]
+  (for [{:keys [name len]} refs]
+    {:chr name :start 1 :end len}))
+
+(defn- sum-quals [a]
+  (when-let [q ^String (:qual a)]
+    (when-not (and (= (.length q) 1) (= \* (.charAt q 0)))
+      (reduce (fn [r x] (+ r (- (int x) 33))) 0 q))))
+
+(defn- paired? [{:keys [flag]}]
+  (pos? (bit-and flag (sam-util/flags :multiple))))
+
+(defn- mapped? [{:keys [flag]}]
+  (zero? (bit-and flag (bit-or (sam-util/flags :unmapped) (sam-util/flags :next-unmapped)))))
+
+(defn dedupe-xform
+  "Returns a transducer which removes PCR duplications."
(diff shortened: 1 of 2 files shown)


src/cljam/dedup.clj
```diff
@@ -0,0 +1,69 @@
...
+              refs->regions
+              (cp/pmap
+               pool
+               (fn [{:keys [start end] :as region}]
+                 (with-open [r (bam/reader in)]
+                   (->> (io/read-alignments r region)
+                        (sequence (dedup-xform :remove-dups remove-dups))
+                        doall))))
+              (sequence cat))
+         header)))))
```
alumi: This might be trivial, but I think `(io/write-alignments writer header alignments)` would be better than `(io/write-alignments writer alignments header)` when used with thread-last macro.

totakke: Ummm... As you say, `(io/write-alignments writer header alignments)` is more useful for `->>`. But `alignments` is main data of this function. `header` is supportive one.

Probably SAMWriter/BAMWriter should has `header` information like `VCFWriter` because `header` is meta data. For example,

```clojure
(with-open [wtr (bam/writer f header)]
  (io/write-alignments wtr alignments)
```

alumi: Thanks. I'll try it in other branch!

alumi:
Actually there's no particular reason but just because I felt that [`dedup`](https://github.com/search?q=dedup+bam&type=Code) is as common as [`dedupe`](https://github.com/search?q=dedupe+bam&type=Code) in the context of bioinformatics. 
OK, let's make it Clojurish. I've modified the commit.

totakke:
Thanks!

PR #34, opened 2017-03-27
Add :ignore-index option to cljam.fasta/reader

totakke:
`cljam.fasta/reader` comes to receive `:ignore-index` option by this commit. (default: `false`)

Current `cljam.fasta/reader` sets `.index` `nil` when an index file is not found. Some functions using index, such as `cljam.fasta/read-sequence`, always returns `nil` if the index is `nil`. This behavior may cause bugs. New `cljam.fasta/reader` disallows nil index unless `:ignore-index true` is passed.


alumi:
Merged 😸

PR #35, opened 2017-04-03
Feature/bam seq encode

alumi:
Refactored bam encoding to use static table.

This PR improves performance of sequence encoding about 20x faster.

Old codes and tests are removed.


src/cljam/util/sam_util.clj
```diff
@@ -161,170 +161,46 @@
...
-                  (char->compressed-base-low (char (aget read-bases l-idx))))]
-          (aset-byte result i (byte (bit-or h l)))
-          (recur (inc i)))))))
-
-(def ^:const ^:private seq-byte-table
-  (let [nibble-table "=ACMGRSVTWYHKDBN"]
-    (->> (for [i nibble-table j nibble-table] [i j])
-         (apply concat)
-         cstr/join)))
+(defonce ^:private ^:const base->nibble "=ACMGRSVTWYHKDBN")
```
alumi: Thank you for the comment.
Fixed the names for static tables with comments to be more descriptive.

As sizes of these static tables are not so trivial (<17KB), I used `defonce` to mark that they shouldn't be reallocated / recalculated.
There're no advantages in runtime in the usual case, as you mentioned. I fixed them to `def`.

totakke:
Thank you!

PR #37, opened 2017-04-03
Add to support lein cloverage

ayamada:
Add `lein cloverage`.

This patch includes two fixes for `lein cloverage`.

- importing `deftype`d classes should need `require` in combination.

- `extend-type` (and `extend`) do `alter-var-root` internally. So `extend-type` should be just below `deftype`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,14 @@ To run heavy tests which uses remote large-size files,
 $ lein midje :filter heavy
 ```
 
+To get coverage, (but it is a half-day job)
+
+```console
+$ lein cloverage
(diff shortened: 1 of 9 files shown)


alumi:
Aside from the code, I think it'd be better if CI could report/log the coverage automatically with [coveralls](https://coveralls.io/) / [codecov](https://codecov.io/) or something. (even better with badges on README.md!)

ayamada:
This PR is not fit for coverage-CI (because incredibly heavy), need another way.

Created #43, and close this.

PR #38, opened 2017-04-04
Fix StackOverflowError in parse-header of SAM/BAM files.

alumi:
This PR fixes a bug in `parse-header` of SAM/BAM files.

Reading SAM/BAM files with too many reference sequences caused StackOverflowError.
This patch replaces explicit recursions with transducers.

### DIFF
--- a/src/cljam/util/sam_util.clj
+++ b/src/cljam/util/sam_util.clj
@@ -11,29 +11,68 @@
 (defn- parse-header-keyvalues
   "e.g. \"LN:45 SN:ref\" -> {:LN 45, :SN \"ref\"}"
   [keyvalues]
-  (apply merge
-         (map (fn [kv]
-                (let [[k v] (cstr/split kv #":")]
-                  {(keyword k) (case k
-                                 "LN" (Integer/parseInt v)
-                                 "PI" (Integer/parseInt v)
-                                 v)}))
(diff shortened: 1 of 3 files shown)


src/cljam/util/sam_util.clj
```diff
@@ -11,29 +11,49 @@
...
+  [rf f]
+  (fn
+    ([] (rf))
+    ([x] (f (rf x)))
+    ([x y] (rf x y))))
+
+(defn into-with-conj
+  "A reducing function that creates a map of vector.
+   e.g. (transduce identity into-with-conj [[:a 1] [:b 2] [:a 3]]) => {:a [1 3] :b [2]}"
+  ([] (transient {}))
```
totakke: `into-with-conj` and `finalize` are not so general. It can be only used in transducers. Non-argument arity of `into-with-conj`, especially, is not good for common use because it returns transient collection. I recommend making these functions private.

alumi: Thank you for the comment.
I split the function `into-with-conj` into `group-by-rf` and `into-rf` to make them a bit more general.
I also made them private and renamed those with prefix `rf` to mark that they return reducing functions which should be used with `transduce`.

totakke:
Thanks!

PR #39, opened 2017-04-06
Add a function to clone bam reader.

alumi:
Reading BAM header and BAI will be time-consuming when BAM has a lot of refs.
Usually we create one reader per ref when process alignments in parallel and reading same BAM header and BAI repeatedly caused degradation of performance.

This PR adds `clone-reader` function to clone BAMReader reconstructing stateful reader objects of Java.


totakke:
Thanks for enhancement and review.

PR #41, opened 2017-04-06
Improves performance of pileup

alumi:
This PR improves performance of pileup. (revised version of #40, this uses feature of #39)

There're two main points.
- use primitive arrays
- make step configurable

Average time of ten profiles piling-up BAM file containing about 65000 alignments in chr1:1-4000000.

[serial]
- before `20918 msecs`
- after `801 msecs`

about 26x faster

[parallel]
- before `8631 msecs`
- after `617 msecs`

about 14x faster

### DIFF
--- a/src/cljam/pileup/pileup.clj
+++ b/src/cljam/pileup/pileup.clj
@@ -2,90 +2,48 @@
   "Provides simple pileup functions."
   (:require [com.climate.claypoole :as cp]
             [cljam.common :refer [get-exec-n-threads]]
+            [cljam.bam :as bam]
             [cljam.util.sam-util :as sam-util]
             [cljam.io :as io]
-            [cljam.pileup.common :refer [window-width step center]]
+            [cljam.pileup.common :as common]
             cljam.bam.reader)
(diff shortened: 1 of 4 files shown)


src/cljam/pileup/pileup.clj
```diff
@@ -2,90 +2,48 @@
...
-             (grouped-rpositions (+ start n) end n))))))
+  [alns beg end]
+  (let [pile (long-array (inc (- end beg)))]
+    (doseq [aln alns]
+      (let [left (max (:pos aln) beg)
+            right (min (sam-util/get-end aln) end)
+            left-index (- left beg)]
+        (dotimes [i (inc (- right left))]
+          (aset-long pile (+ i left-index) (inc (aget pile (+ i left-index)))))))
+    (seq pile)))
```
totakke: New `count-for-positions` returns a result as a sequence. Please fix docstring of `count-for-positions`.

alumi: Thank you for the review. I fixed the docstring and rebased :bowing_man:.

totakke:
Thank you.

PR #48, opened 2017-04-18
Use lein-binplus instead of lein-bin

totakke:
[lein-bin](https://github.com/Raynes/lein-bin) plugin is not been maintained anymore. [lein-binplus](https://github.com/BrunoBonacci/lein-binplus) is active fork of lein-bin.

Additionally,

* `:bin {:bootclasspath true}` for launch performance
* `version` sub-command for printing version

were added.


alumi:
Thank you :+1: merged & deployed.

PR #49, opened 2017-04-18
Don't throw exception in `delay`

ayamada:
Evaluate `(bam-index f)` in `delay`, but cannot throw exception in `delay`.
`deref` this to get just a exception instance.
It is not desirable result.

Instead, put mere `nil` value.

### DIFF
--- a/src/cljam/bam/core.clj
+++ b/src/cljam/bam/core.clj
@@ -19,11 +19,10 @@
 
 (defn- bam-index [f & {:keys [ignore]
                        :or {ignore false}}]
-  (if-not ignore
+  (when-not ignore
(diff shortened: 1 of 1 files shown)


alumi:
I don't get the point.
> `deref` this to get just a exception instance.

I think `throw` in `deref` still causes exception.
https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/Delay.java#L46

An IOException is thrown if

- BAI doesn't exists
- Operations which need index are invoked

and this is reasonable.

totakke:
I think so, too. The original behavior seems to be right.

ayamada:
Sorry, it is my mistake... thanks.

PR #51, opened 2017-04-18
Refine unused code for newline

ayamada:
`newline-character`, `return-character`, `newline*?` and `newline?` were not used, refine it.

### DIFF
--- a/src/cljam/fasta/reader.clj
+++ b/src/cljam/fasta/reader.clj
@@ -121,8 +121,8 @@
          (.clear ~buffer)
          ba#)))
 
-(def ^:private gt-byte (byte \>))
-(def ^:private newline-byte (byte \newline))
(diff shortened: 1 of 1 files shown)


src/cljam/fasta/reader.clj
```diff
@@ -131,6 +121,9 @@
          (.clear ~buffer)
          ba#)))
 
+(def ^:private gt-byte (byte \>))
+(def ^:private newline-byte (byte \newline))
```
totakke: Please add `^:const` to these vars for equivalence of the original code.

ayamada: Thanks to your review. I fixed it by https://github.com/chrovis/cljam/pull/51/commits/3c4fcdd4574e3cd70e130d30e50cffe3369369d6 .

PR #53, opened 2017-04-18
Fix single-use reader for n-threads = 1

ayamada:
Need to clone to reader not only multi-thread but also single-thead.
This threw exception without cloning.

And add related tests.

### DIFF
--- a/src/cljam/pileup/pileup.clj
+++ b/src/cljam/pileup/pileup.clj
@@ -36,7 +36,8 @@
         count-fn (fn [xs]
                    (if (= n-threads 1)
                      (map (fn [[start end]]
-                            (count-for-positions (read-fn rdr start end) start end)) xs)
+                            (with-open [r (bam/clone-reader rdr)]
+                              (count-for-positions (read-fn r start end) start end))) xs)
                      (cp/pmap (dec n-threads)
                               (fn [[start end]]
                                 (with-open [r (bam/clone-reader rdr)]
--- a/test/cljam/t_pileup.clj
+++ b/test/cljam/t_pileup.clj
@@ -42,26 +42,28 @@
   (is (= (with-open [r (bam/reader test-sorted-bam-file)]
            (plp/pileup r "ref" nil))
          test-bam-pileup-ref))
-  (is (= (with-open [r (bam/reader test-sorted-bam-file)]
-           (plp/pileup r "ref" {:step 2 :n-threads 4}))
-         test-bam-pileup-ref))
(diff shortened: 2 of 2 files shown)


src/cljam/pileup/pileup.clj
```diff
@@ -36,7 +36,8 @@
         count-fn (fn [xs]
                    (if (= n-threads 1)
                      (map (fn [[start end]]
-                            (count-for-positions (read-fn rdr start end) start end)) xs)
+                            (with-open [r (bam/clone-reader rdr)]
```
alumi: We don't need to clone the reader because it'll be accessed from only one thread.

test/cljam/t_pileup.clj
```diff
@@ -42,26 +42,28 @@
...
-           (plp/pileup r "ref" {:step 11 :n-threads 4}))
-         test-bam-pileup-ref))
-  (is (= (with-open [r (bam/reader test-sorted-bam-file)]
-           (plp/pileup r "ref" {:step 13 :n-threads 4}))
-         test-bam-pileup-ref))
-  (is (= (with-open [r (bam/reader test-sorted-bam-file)]
-           (plp/pileup r "ref" {:n-threads 4}) test-bam-pileup-ref)))
+  (doseq [n-threads [4 1]]
+    (is (= (with-open [r (bam/reader test-sorted-bam-file)]
+             (plp/pileup r "ref" {:step 2 :n-threads n-threads}))
```
alumi: Instead of `clone-reader`, we need `doall` here to finish piling-up before closing the reader.

test/cljam/t_pileup.clj
```diff
@@ -42,26 +42,28 @@
...
+           test-bam-pileup-ref))
+    (is (= (with-open [r (bam/reader test-sorted-bam-file)]
+             (plp/pileup r "ref" {:step 11 :n-threads n-threads}))
+           test-bam-pileup-ref))
+    (is (= (with-open [r (bam/reader test-sorted-bam-file)]
+             (plp/pileup r "ref" {:step 13 :n-threads n-threads}))
+           test-bam-pileup-ref))
+    (is (= (with-open [r (bam/reader test-sorted-bam-file)]
+             (plp/pileup r "ref" {:n-threads n-threads})
+             test-bam-pileup-ref))))
```
alumi: I'm sorry, the original test case was incorrect. `test-bam-pileup-ref` must be outside of `with-open`

ayamada:
@alumi Thanks to your review.
I understand to this PR is not correct. I close this.
Fixes for test will be included in another PR at later.

PR #54, opened 2017-04-18
Pileup without options

ayamada:


src/cljam/pileup.clj
```diff
@@ -15,6 +15,8 @@
   {:n-threads 0})
 
 (defn pileup
+  ([bam-reader rname]
+   (plp/pileup bam-reader rname))
```
alumi: `(pileup bam-reader rname {})` would be better, I think.

ayamada: Thanks to your review.
I fixed it by https://github.com/chrovis/cljam/pull/54/commits/64eeb38b8fcbe71eae014a1e8c012d0902ecbae0 .

PR #58, opened 2017-04-18
Activate cljam.fasta/slurp

ayamada:
`cljam.fasta.core/slurp` exists, but `cljam.fasta/slurp` don't exists.
Create it and add tests.

If should not to go public it, please don't merge this PR and close this.
(I cannot judge this)


totakke:
No, `cljam.fasta.core/slurp` is not necessary any more. Please remove it.

PR #59, opened 2017-04-18
Activate cljam.fasta/sequential-read-byte-array

ayamada:
`cljam.fasta.reader/sequential-read-byte-array` exists, but don't used.
Add `cljam.fasta/sequential-read-byte-array` to call this, and add related tests.

If should not to go public it, please don't merge this PR and close this.
(I cannot judge this)

### DIFF
--- a/src/cljam/fasta.clj
+++ b/src/cljam/fasta.clj
@@ -45,6 +45,15 @@
   [rdr]
   (fa-core/reset rdr))
 
+(defn sequential-read-byte-array
+  "Reads entire sequences sequentially on caller's thread,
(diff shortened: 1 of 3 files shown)


alumi:
I use this function from another project.
As the return value format (`1,2,3,4,5` for `ATGCN`) is not used in other part of `cljam`, I think it shouldn't be in `cljam.fasta`.
I want to leave it as is.

totakke:
This function is not easy to be used by public users. No need to make it public.

ayamada:
So, I understood it. Thanks!

Issue #61, opened 2017-04-20
Add tests to check generated data correctness


ayamada:
(Came from https://github.com/chrovis/cljam/pull/60#pullrequestreview-33452728 )

Add many tests by #60, but some tests doesn't check generated(converted) result data.
Should check these data (but may need correct result data).

1. https://github.com/chrovis/cljam/blob/ff91c66449ddd40b9a48355531550b12fe42c92b/test/cljam/t_cli.clj#L53
    - [x] Done by #66

2. https://github.com/chrovis/cljam/blob/ff91c66449ddd40b9a48355531550b12fe42c92b/test/cljam/t_cli.clj#L96
    - [x] Done by #65
    - But a comment say `NB: "pileup" output format may change in future (maybe)`.

3. https://github.com/chrovis/cljam/blob/ff91c66449ddd40b9a48355531550b12fe42c92b/test/cljam/t_cli.clj#L142
    - [x] Done by #70

4. https://github.com/chrovis/cljam/blob/ff91c66449ddd40b9a48355531550b12fe42c92b/test/cljam/t_pileup.clj#L218
    - [x] Done by #67

5. https://github.com/chrovis/cljam/blob/ff91c66449ddd40b9a48355531550b12fe42c92b/test/cljam/t_dict.clj#L17
    - [x] Done by #64

6. https://github.com/chrovis/cljam/blob/ff91c66449ddd40b9a48355531550b12fe42c92b/test/cljam/t_dedupe.clj#L11
    - [x] Done by #63

7. https://github.com/chrovis/cljam/blob/ff91c66449ddd40b9a48355531550b12fe42c92b/test/cljam/t_sorter.clj#L110
    - [ ] Need lightweight bam-file with many blocks (require to over [1500000](https://github.com/chrovis/cljam/blob/ff91c66449ddd40b9a48355531550b12fe42c92b/src/cljam/sorter.clj#L14)) for sorter test coverage.
    - I commit https://github.com/ayamada/cljam/commit/f2ed61ba6b52b7d4bd44abf6c653c20179278b2a , it can work, but it cause very long time in `lein cloverage`, may not merge it.

ayamada:
I numbered items of this issue.

Seems (1)(2)(3)(4)(5) can be assigned to current `cljam` result for controlled test result.
(But it is very simple tests only comparing files, not perfect.)

Shall I create PR to compare contorolled test result for test (1)(2)(3)(4)(5) ?

ayamada:
[...]

PR #63, opened 2017-04-21
Issue #61 (6) Examine result file of `dedupe` in test

ayamada:


ayamada:
Thanks for review, I changes to new bam files based on another dedupe test.
Is it OK?

totakke:
Great:) Thank you.

PR #64, opened 2017-04-21
Issue #61 (5) Examine result file of `create-dict` in test

ayamada:


test/cljam/t_dict.clj
```diff
@@ -2,9 +2,26 @@
...
+          ;; NB: `UR` is calculated by file path, ignore it
+          omit-ur (fn [line]
+                    (if-let [[_ r] (re-find #"^(.*)\tUR:" line)]
+                      r
+                      line))]
+      (when (= (count dict1) (count dict2))
+        (->> (map #(= (omit-ur %1) (omit-ur %2)) dict1 dict2)
+             (filter identity)
+             first
+             boolean)))))
```
totakke: Is this implementation right? This seems to return `false` only if all lines are different. (`true` if any pair of lines is equal.)

Is the following enough?

```clojure
(->> (map #(= (omit-ur %1) (omit-ur %2)) dict1 dict2)
     (every? true?))
```

Additionally, `VN:1.4` can be ignore. `omit-vn` is required.

ayamada: Thanks for your review, It is my mistake.
I fixed these problems https://github.com/ayamada/cljam/commit/507f8651d82bb61a5810fcc806115012da13cad4 .

totakke:
Thanks!

PR #72, opened 2017-05-01
Separate :slow and :heavy test-selectors

totakke:
`:slow` and `:heavy` tests are substantially same on current test-selectors because all `:heavy` tests also have `:slow` tag. It is difficult to run only slow tests with local resources.

`:slow` and `:heavy` tests are clearly separated by this PR.

* `lein test :slow` to run slow tests with local resources.
* `lein test :heavy` to run tests with remote resources.
* `lein test :slow :heavy` to run slow tests with local and remote resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,17 +128,11 @@ $ cljam view --header path/to/file.sam
 
 ### Test
 
-To run all basic tests,
+To run tests,
 
-```console
-$ lein test
-```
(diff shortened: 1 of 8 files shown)


totakke:
You're right. I'll change `:heavy` to `:remote` before merge. Thanks.

PR #73, opened 2017-05-09
BED reader/writer

alumi:
Added reader/writer for BED file format.
We can now read/write BED file compressed with gzip and bzip2.


src/cljam/bed.clj
```diff
@@ -67,6 +86,25 @@
 (defn- serialize-bed
   "Serialize bed fields into string."
   [m]
+  {:pre [(and (:chr m) (:start m) (:end m))
```
totakke: `and` is not needed because all items in `:pre` vector are combined with `and`.

src/cljam/bed.clj
```diff
@@ -67,6 +86,25 @@
...
+         (if-let [xs (:block-sizes m)] (= (count xs) (:block-count m)) true)
+         ;; The number of items in this list should correspond to blockCount.
+         (if-let [xs (:block-starts m)] (= (count xs) (:block-count m)) true)
+         ;; The number of items in this list should correspond to blockCount.
+         (if-let [[f] (:block-starts m)] (= 0 f) true)
+         ;; The first blockStart value must be 0.
+         (if-let [xs (:block-starts m)] (= (+ (last xs) (last (:block-sizes m))) (- (:end m) (:start m))) true)
+         ;; The final blockStart position plus the final blockSize value must equal chromEnd.
+         (if-let [xs (:block-starts m)] (apply <= (mapcat (fn [a b] [a (+ a b)]) xs (:block-sizes m))) true)
+         ;; Blocks may not overlap.
```
totakke: I think a comment should be placed just before a related predicate.

```clojure
{:pre [;; First 3 fields are required.
       (:chr m) (:start m) (:end m)
       ;; The chromEnd base is not included in the display of the feature.
       (< (:start m) (:end m))
       ...
       ]}
```

src/cljam/bed.clj
```diff
@@ -1,8 +1,27 @@
...
+  Closeable
+  (close [this]
+    (.close ^Closeable (.reader this))))
+
+(defrecord BEDWriter [^BufferedWriter writer ^String f]
+  Closeable
+  (close [this]
+    (.close ^Closeable (.writer this))))
+
+(defn ^BEDReader reader [f]
```
totakke: Please add docstring. Also `writer`.

alumi:
Thank you for the review! I've added a commit.

PR #75, opened 2017-05-15
Extend cljam.core reader/writer functions.

alumi:
Current `cljam.core` supports only SAM, BAM and FASTA files.

This PR adds other readers/writers to `cljam.core`.


alumi:
Thank you for the reviewing!

PR #77, opened 2017-05-16
Improve performance of deep decoding of BAM files.

alumi:
This PR adds some performance optimization for reading BAM files.

- Use `defrecord` for alignment instead of map
- Avoid multimethod
- Parse options in `lazy-seq`
- Avoid char in `case`
- Compute `:end` while decoding cigar code and cache them
- Read little-endian values directly from `ByteBuffer`.

Sequential reading will get **50~60% faster** when `:options` are not referred and **30~40% faster** with `:options`.


src/cljam/bam/decoder.clj
```diff
@@ -87,61 +92,55 @@
...
-          qname       (lsb/read-string buffer (dec l-read-name))
-          _           (lsb/skip buffer 1)
-          cigar       (decode-cigar (lsb/read-bytes buffer (* n-cigar-op 4)))
-          seq         (decode-seq (lsb/read-bytes buffer (quot (inc l-seq) 2)) l-seq)
-          qual        (decode-qual (lsb/read-bytes buffer l-seq))
-          rest        (lsb/read-bytes buffer (options-size (:size block)
-                                                           l-read-name
-                                                           n-cigar-op
-                                                           l-seq))
+          pos         (inc (.getInt buffer))
```
totakke: These changes about buffer reading reduces readability and component merit. I think enhancing each `lsb/read-***` function with protocol is better. If so, performance of `cljam.bcf.reader` will improve as well.

src/cljam/bam/decoder.clj
```diff
@@ -87,61 +92,55 @@
...
-                          (.append (decode-cigar-op op)))))
-        (str sb)))))
+          (doto sb
+            (.append n)
+            (.append (case op 0 \M 1 \I 2 \D 3 \N 4 \S 5 \H 6 \P 7 \= 8 \X)))
+          (recur (+ ref-length (case op 0 n 2 n 3 n 7 n 8 n 0))))
+        [(.toString sb) ref-length]))))
+
+(defrecord SAMAlignment [^String qname ^int flag ^String rname ^int pos ^int end ^int mapq ^String cigar
+                         ^String rnext ^int pnext ^int tlen ^String seq ^String qual options])
```
alumi: I changed `cljam.sam.reader/read-alignments` to return `SAMAlignment`s.
`SAMAlignment` is now defined in `cljam.io` without `^String` type hints.

src/cljam/bam/decoder.clj
alumi: I replaced `lsb/read-***` functions with a protocol and extensions.

src/cljam/util/sam_util.clj
```diff
@@ -330,8 +334,8 @@
 (defn ref-name
   "Returns a reference name from the reference ID. Returns nil if id is not
   mapped."
-  [refs id]
-  (if (<= 0 id (dec (count refs)))
+  [refs ^long id]
+  (when (<= 0 id (dec (count refs)))
```
alumi: I prefer `when` because one can tell at a glance that the function may return `nil`.

In [clojure.core.clj](https://github.com/clojure/clojure/blob/master/src/clj/clojure/core.clj), `if`s without third arg are less than 1% of entire use of `if`, while `when` with  single body is commonly used.

alumi:
Thank you for the reviewing!
I've added some more commits to resolve problems.

src/cljam/util/sam_util.clj
totakke: I personally think existence of side effects is more important than whether the function return `nil`. The third argument of `if` is not a serious matter. Well, OK in this case because style guide about `if/when` hasn't been established yet.

totakke:
Thank you for fixes :)

alumi:
Thank you so much!

PR #78, opened 2017-05-26
Add protocols for I/O APIs.

alumi:
**This PR includes breaking changes**

#### Summary
This PR adds new protocols for I/O classes for the sake of
- Abstraction of concrete implementation of file formats
- Improving consistency of APIs
- Convenience for REPL use

#### Changes
- Add protocols to `cljam.io`
- Modify readers/writers to conform to the protocols
- Modify arguments to take region map like `{:chr "chr1", :start 1, :end 3}` and option as another map if required.
- Add util function for manipulating region map.
- Use unique aliases: `cio` for `clojure.java.io` and `io` for `cljam.io`
- Replace vararg `[& {:keys [...]}]` with map `[{:keys [...]}]`
- Arity overloadings default to call same fn with larger arity

#### Problems
`read-in-region` reads contents in certain region.
In current implementation, readers without index (SAM, VCF...) are still supporting the function by `filter`.
Even though they will warn about performance, I'm still thinking if we should disable this.

### DIFF
--- a/src/cljam/io.clj
+++ b/src/cljam/io.clj
@@ -1,25 +1,56 @@
 (ns cljam.io
-  "Protocols of SAM/BAM reader.")
+  "Protocols of reader/writer for various file formats."
+  (:refer-clojure :exclude [read]))
 
 (defrecord SAMAlignment
   [qname ^int flag rname ^int pos ^int end ^int mapq cigar rnext ^int pnext ^int tlen seq qual options])
 
-(defprotocol ISAMReader
-  (reader-path [this] "Returns the file's absolute path.")
-  (read-header [this] "Returns header of the SAM/BAM file.")
-  (read-refs [this] "Returns references of the SAM/BAM file.")
-  (read-alignments [this] [this option]
-    "Reads alignments of the SAM/BAM file, returning the alignments as a lazy
-    sequence.")
-  (read-blocks [this] [this option]
-    "Reads alignment blocks of the SAM/BAM file, returning the blocks as a lazy
-    sequence."))
+(defprotocol IReader
+  (reader-path [this]
+    "Returns the file's absolute path.")
+  (read [this] [this option]
+    "Sequentially reads contents of the file."))
 
-(defprotocol ISAMWriter
-  (writer-path [this] "Returns the file's absolute path.")
-  (write-header [this header] "Writes header to the SAM/BAM file.")
-  (write-refs [this header] "Writes references to the SAM/BAM file.")
+(defprotocol IWriter
+  (writer-path [this]
+    "Returns the file's absolute path."))
+
+(defprotocol IRegionReader
+  (read-in-region [this region] [this region option]
+    "Reads contents of the file in given region."))
+
+(defprotocol IAlignmentReader
+  (read-header [this]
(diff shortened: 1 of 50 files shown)


src/cljam/io.clj
```diff
@@ -1,25 +1,56 @@
...
-
-(defprotocol ISAMWriter
-  (writer-path [this] "Returns the file's absolute path.")
-  (write-header [this header] "Writes header to the SAM/BAM file.")
-  (write-refs [this header] "Writes references to the SAM/BAM file.")
+(defprotocol IReader
+  (reader-path [this]
+    "Returns the file's absolute path.")
+  (read [this] [this option]
+    "Sequentially reads cotents of the file."))
```
totakke: `cotents` -> `contents`

src/cljam/io.clj
```diff
@@ -1,25 +1,56 @@
...
+
+(defprotocol IWriter
+  (writer-path [this]
+    "Returns the file's absolute path."))
+
+(defprotocol IRandomReader
+  (read-in-region [this {:keys [chr start end]}] [this {:keys [chr start end]} option]
```
totakke: Please use `region` consistently instead of `{:keys [chr start end]}` in protocol definition.

src/cljam/util.clj
```diff
@@ -149,8 +149,30 @@
...
        (-> (CompressorStreamFactory.)
            (.createCompressorOutputStream s os))
        os))))
+
+;; region utils
+;; ---------
+
+(defn divide-region
+  "Divides a region [start end] into several chunks with maximum length 'step'.
+   Returns a lazy sequence of vector."
```
totakke: Indent of the second line is incorrect.

```clojure
(defn divide-region
  "Divides a region [start end] into several chunks with maximum length 'step'.
  Returns a lazy sequence of vector."
  ...)
```

is correct.

c.f. [clojure-style-guide#align-docstring-lines](https://github.com/bbatsov/clojure-style-guide#align-docstring-lines)

totakke:
> read-in-region reads contents in certain region.
> In current implementation, readers without index (SAM, VCF...) are still supporting the function by filter.
> Even though they will warn about performance, I'm still thinking if we should disable this.

Readers w/o index should not implement `IRandomReader` because they are not suitable for random access.  I think the following solutions:

1. Renaming `IRandomReader` to `IRegionReader`.
    * `IRegionReader` does not mention random access.
2. Providing `IRegionReader` and `IRandomRegionReader`.
    * Readers with index implement `IRandomRegionReader` and reader w/o index implement `IRegionReader`. If so, users can detect them with `satisfies?`.
3. Readers w/o index do not implement `IRandomReader`.

totakke:
Oh, I forgot to add one comment. Please unstage change of `README.md`. I will fix `README.md` when releasing new version.

alumi:
Thank you for your review!
I've fixed the docs and the arg-list.

For `read-in-region`, I've chosen the first option.

> 1. Renaming IRandomReader to IRegionReader.
>   IRegionReader does not mention random access.

Commits are rebased so that `README.md` is not changed now.

totakke:
Thanks! Please wait merge until v0.3.1 release because of breaking changes.

PR #80, opened 2017-05-30
Remove dependency on raynes/fs

totakke:
I removed dependency on [raynes/fs](https://github.com/Raynes/fs) and fixed related codes because raynes/fs might not be maintained any more.


alumi:
Thanks!
Is there any problem with upgrading cavia 0.4.0 to 0.4.1?

alumi:
Sorry🙇 , cavia 0.4.1 is not released yet.

totakke:
Thanks for merge. I'll release cavia 0.4.1 by the next cljam release.

PR #84, opened 2017-06-21
I/O API refactoring

totakke:
Breaking refactoring of I/O API.

Old `cljam.io` is not user-friendly, moving to `cljam.io.protocols`. Basically, this namespace should not be used directly on other projects. It can be used in cljam project or for advanced usage.

Public namespaces related to a format and its binary equivalent are integrated. For example,

* `cljam.io.sam` supports SAM and BAM,
* `cljam.io.vcf` supports VCF and BCF,
* `cljam.io.sequence` supports FASTA and TwoBit.

These have read/write functions wrapping `cljam.io.protocols`. Only commonly-used functions are wrapped. A user just `require` a single namespace that handles a format.

```clojure
(require '[cljam.io.sam :as sam])

(with-open [rdr (sam/reader "foo.bam")]
  (sam/read-alignments rdr))
```


src/cljam/io/util.clj
```diff
@@ -136,38 +140,3 @@
     #"(?i)\.bcf$" :bcf
     #"(?i)\.bed" :bed
     (throw (IllegalArgumentException. "Invalid file type"))))
-
-(defn ^Closeable reader
```
alumi: How about leaving `reader/writer` functions here?
It's true that they're unnecessary for developing apps, but I think those versatile functions are still useful for REPL.

totakke: That causes cyclic load. `cljam.sam` loads `cljam.io.util` for `cljam.sam/reader`, and `cljam.io.util` loads `cljam.sam` for `cljam.io.util/reader`. If `reader/writer` are realy necessary, I have to reconsider namespace allocation.

alumi: Thank you for pointing that out. OK, Let's remove them.

alumi:
Merged. Thank you! 👍

totakke:
Thank you for review!

PR #85, opened 2017-06-22
Fix commented tests

totakke:
I forgot to fix commented tests of predicates and old `reader/writer` in `cljam.io.util`. This PR fixes them.


alumi:
Merged. Thank you for the fixes!

PR #87, opened 2017-06-26
Use proton instead of some utilities

totakke:
Some utility functions, such as `str->int`, are replaced by [proton](https://github.com/xcoo/proton)'s equivalences.

Additionally, deprecated `String.getBytes(int,int,byte[],int)` method is replaced by `String.getBytes()`.

`cljam.util/swap` macro is removed since it is used in only `cljam.algo.normal` and it can be rewritten with `update` as function.


alumi:
Merged. Thanks!

PR #92, opened 2017-07-31
Fix parser of optional fields in SAM/BAM.

alumi:
Fixed returned type of following two elements in optional fields in SAM/BAM.

|type|description|module|type (before)|type (after)|
|:-:|:-:|:-:|:-:|:-:|
|A|Printable character|bam|byte|char|
|H|Byte array in the Hex format|sam|nil|byte-array|


src/cljam/io/sam/util.clj
```diff
@@ -77,6 +77,19 @@
         (map parse-header-line)
         into-header)))
 
+(defn- ^"[B" as-byte-array [^String s]
```
totakke: proton has `hex->bytes` function, which seems to be equivalent of this function. Is any reason to reinvent it?

alumi: I hadn't noticed that fn. Thank you for pointing that!

totakke:
Thank you :)

alumi:
Thank you for the reviewing!

PR #94, opened 2017-08-04
Update an example usage of depth in README.

alumi:
Sorry, I didn't update README when I renamed `cljam.algo.pileup` to `cljam.algo.depth`.


alumi:
Thanks for your review! 😺

PR #96, opened 2017-08-07
Refactor BAI I/O

alumi:
#### summary

Overall refactoring of BAI module, including performance improvements.

#### changes

- Use the linear index to reduce spans.
    - For some reason, the linear index was not used to optimize chunks.
    - > Given a region [rbeg,rend), we only need to visit a chunk whose end file offset is larger than the file offset of the 16kbp window containing rbeg.
- Search more BAI files for a BAM file.
    - Given `"foo.bam"`, `["foo.bam.bai", "foo.bai", "foo.bam.BAI", "foo.BAI"]` are now looked up.
- Use records for chunks.
- Use ref-index instead of rname in bam index writer.
- Remove unused local vars.

Thanks.

### DIFF
--- a/src/cljam/io/bam/core.clj
+++ b/src/cljam/io/bam/core.clj
@@ -18,27 +18,25 @@
 
 (defn- bam-index
   "Load an index file (BAI) for the given BAM file."
-  [bam-path {:keys [ignore]}]
-  (let [bai-path (->> ["$1.bai" ".bai" "$1.BAI" ".BAI"]
-                      (eduction
-                       (comp
-                        (map #(cstr/replace bam-path #"(?i)(\.bam)$" %))
-                        (filter #(.isFile (cio/file %)))))
-                      first)]
-    (if (and bai-path (not ignore))
-      (bai/bam-index bai-path)
-      (throw (FileNotFoundException. (str "Could not find BAM Index file for " bam-path))))))
+  [bam-path]
+  (if-let [bai-path (->> ["$1.bai" ".bai" "$1.BAI" ".BAI"]
+                         (eduction
+                          (comp
+                           (map #(cstr/replace bam-path #"(?i)(\.bam)$" %))
+                           (filter #(.isFile (cio/file %)))))
+                         first)]
+    (bai/bam-index bai-path)
+    (throw (FileNotFoundException. (str "Could not find BAM Index file for " bam-path)))))
 
 (defn ^BAMReader reader
-  "Creates a `cljam.io.bam.BAMReader` instancer for the given path.
-  Pass {:ignore-index true} to force to ignore BAI file."
-  [f {:keys [ignore-index]}]
+  "Creates a `cljam.io.bam.BAMReader` instance for the given path."
+  [f]
   (let [rdr (BGZFInputStream. (cio/file f))
         data-rdr (DataInputStream. rdr)]
     (when-not (Arrays/equals ^bytes (lsb/read-bytes data-rdr 4) (.getBytes ^String bam-magic))
       (throw (IOException. "Invalid BAM file")))
     (let [{:keys [header refs]} (reader/load-headers data-rdr)
-          index-delay (delay (bam-index f {:ignore ignore-index}))]
+          index-delay (delay (bam-index f))]
       (BAMReader. (.getAbsolutePath (cio/file f))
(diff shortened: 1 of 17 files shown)


src/cljam/io/bam_index/writer.clj
```diff
@@ -76,11 +77,11 @@
 (defn- update-bin-index
   [bin-index aln]
   (let [bin (sam-util/compute-bin aln)
-        achunk (:chunk (:meta aln))]
+        achunk ^Chunk (chunk/map->Chunk (:chunk (:meta aln)))]
```
totakke: Record's `map->Foo` function is slow. I recommend using `->Foo` function or Java constructor `Foo.`.

src/cljam/io/bam_index/writer.clj
```diff
@@ -4,26 +4,27 @@
             [cljam.io.sam.util :as sam-util]
             [cljam.io.util.bgzf :as bgzf]
             [cljam.io.util.lsb :as lsb]
-            [cljam.io.bam.decoder :as bam-decoder]
             [cljam.io.bam-index.common :refer :all]
-            [cljam.io.bam-index.chunk :as chunk])
-  (:import [java.nio ByteBuffer ByteOrder]))
+            [cljam.io.bam-index.chunk :as chunk]
+            [cljam.io.bam.decoder])
```
alumi: I noticed that the decoder was used to perform decoding in parallel for better performance.
I restored the decoding part.
Thank you.

src/cljam/io/bam_index/writer.clj
alumi: Thanks for your advice. I replaced it with `Chunk.`.

src/cljam/io/bam/core.clj
```diff
@@ -1,36 +1,44 @@
...
-      (if (.isFile (cio/file bai-f))
-        (bai/bam-index bai-f)
-        (throw (IOException. "Could not find BAM Index file"))))))
+(defn- bam-index
+  "Load an index file (BAI) for the given BAM file."
+  [bam-path & [{:keys [ignore]}]]
+  (let [bai-path (->> ["$1.bai" ".bai" "$1.BAI" ".BAI"]
+                      (eduction
+                       (comp
+                        (map #(cstr/replace bam-path #"(?i)(\.bam)" %))
```
alumi: Thanks for pointing that out!  I didn't notice that. 🤕

src/cljam/io/bam/core.clj
```diff
@@ -1,36 +1,44 @@
...
+                        (filter #(.isFile (cio/file %)))))
+                      first)]
+    (if (and bai-path (not ignore))
```
alumi: Thanks for your comment.
As you said, I've changed the behavior from the master version in this PR.

The `:ignore` option was originally introduced to skip loading an index file before constructing BAMReader.
But because `cljam.io.bam.core/bam-index` is now always called in `delay`,
the actual loading of the index file (or throwing an exception) will happen when the first time a random access is queried.
So it's just a matter of the position of throwing exceptions:
- master: nil-handling part in BAM reader
- this PR: inside of the bam-index function

And I think the latter is simpler and more consistent.

alumi:
Thank you for reviewing!
I've updated the commit.

I have a question about ignoring bai: https://github.com/chrovis/cljam/pull/96#discussion_r131858258
I'd appreciate your reply.

src/cljam/io/bam/core.clj
totakke: I understand your intention and agree to the change. But if so, `:ignore` option seems not to be needed. Is it left for compatibility?

alumi: Yes, it's been only a kind of assertion for a while...

I think it's not so hard to handle the compatibility issues on my other products.
I'll remove the options if it's okay with you 🤓

totakke: I think there is no problem. Please go ahead. 😄

alumi: I added a commit to remove the `:ignore` option and I confirmed that it passes `test :all`.

totakke:
LGTM. Thank you!

alumi:
I appreciate your helpful comments 😄

PR #98, opened 2017-08-18
Tweak the normalization function for chromosome name

federkasten:
This PR adds following features,

- Parse a suffix which represents contig type and version (e.g. `v2_alt`)
- Keep that suffix in lower case to work with reference files that are available from UCSC
- Add a function to determine whether chromosome name is a primary one

### DIFF
--- a/src/cljam/util/chromosome.clj
+++ b/src/cljam/util/chromosome.clj
@@ -45,7 +45,7 @@
              normalize-chromosome-prefix)
          (if version-suffix (cstr/lower-case version-suffix)))))
 
-(defn is-primary?
+(defn is-primary-chromosome?
   [s]
   (not (nil? (re-matches #"^chr([0-9]{1,2}|X|Y|M|MT)$"
                          (normalize-chromosome-key s)))))
--- a/src/cljam/util/chromosome.clj
+++ b/src/cljam/util/chromosome.clj
@@ -13,6 +13,11 @@
   [s]
   (cstr/replace s #"(?i)^chr" ""))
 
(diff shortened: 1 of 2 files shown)


src/cljam/util/chromosome.clj
```diff
@@ -13,6 +13,11 @@
   [s]
   (cstr/replace s #"(?i)^chr" ""))
 
+(defn- split-version-suffix
+  [s]
+  (let [[_ base suffix leftover] (re-matches #"(.+?)((?i)v[0-9](_alt|_random)?)?$" s)]
```
totakke: `leftover`  is never used, please replace it with `_`. `$` is not needed for `re-matches`.

src/cljam/util/chromosome.clj
```diff
@@ -32,7 +37,14 @@
...
+  (let [[base version-suffix] (split-version-suffix s)]
+    (str (-> base
+             normalize-name
+             prepend-chromosome-prefix
+             normalize-chromosome-prefix)
+         (if version-suffix (cstr/lower-case version-suffix)))))
+
+(defn is-primary-chromosome?
+  [s]
+  (not (nil? (re-matches #"^chr([0-9]{1,2}|X|Y|M|MT)$"
```
totakke: `not nil?` can be replaced with `some?`, and `^$` is not needed for `re-matches`.

federkasten:
Thanks for reviewing! I have fixed the points.

totakke:
Thank you!

PR #100, opened 2017-08-30
Make index reading of FASTA reader delayed

totakke:
for consistency with https://github.com/chrovis/cljam/pull/96#discussion_r131858258. `:ignore-index` option is removed.


alumi:
Thank you! LGTM 😸

PR #102, opened 2017-08-31
Use lazy pmap for lazy-depth

totakke:
`com.climate.claypoole/pmap` consumes extra elements, so that it is inappropriate for lazy sequence. Instead, `cljam.algo.depth/lazy-depth` uses `com.climate.claypoole.lazy/pmap` by this change.

`lazy/pmap` is inferior in speed to `cp/pmap`, but current cljam provides non-lazy `depth` function for performance. I think `lazy/pmap` is useful to handle returned lazy sequence of `lazy-depth`.


alumi:
You're absolutely right! Thank you!

PR #105, opened 2017-09-04
Rename test namespaces to conventional names

totakke:
Renames test namespaces to conventional names of clojure.test. e.g. `cljam.util.t-chromosome` -> `cljam.util.chromosome-test`

[cider-test](http://cider.readthedocs.io/en/latest/running_tests/) infers a test namespace by adding a suffix, `-test`, so `t-some` namespace cannot be run by `cider-test-run-test`. The inference can be customized in local config, but I feel `some-test` style is better for clojure.test.


alumi:
Great! 👍
I checked `cider-test-run-ns-tests` and `cider-auto-test-mode` working fine with the updated ns.

totakke:
Thanks for the confirmation!

PR #109, opened 2017-09-08
Add `read-indices` to ISequenceReader protocol.

alumi:
#### Summary
Add `read-indices` to ISequenceReader protocol, which was defined only for FASTA but not for 2bit.
Also defined an alias function in `cljam.io.sequence`.

#### Tests

- `lein test :all` 🆗


alumi:
Thanks! 😸

PR #111, opened 2017-09-19
Fix parsing comma-separated number in region

totakke:
Adds support of comma-separated number in region.

```clojure
(region/parse-region "chr1:1,000-2,000")
;;=> {:chr "chr1", :start 1000, :end 2000}
```

SAMtools also supports such numbers.

```
5. A region should be presented in one of the following formats:
   `chr1', `chr2:1,000' and `chr3:1000-2,000'. When a region is
   specified, the input alignment file must be a sorted and indexed
   alignment (BAM/CRAM) file.
```

### DIFF
--- a/src/cljam/util/region.clj
+++ b/src/cljam/util/region.clj
@@ -127,7 +127,7 @@
   "Parse a region string into a map."
   [region-str]
   (when region-str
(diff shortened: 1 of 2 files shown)


alumi:
Merged. Thank you for the fix 👍

PR #112, opened 2017-09-27
Add region option to view command

totakke:
Adds region option (`--region`, `-r`) to `view` command.

```console
$ lein run view -r ref2:1-9 test-resources/bam/test.sorted.bam
x1      0       ref2    1       30      20M     *       0       0       AGGTTTTATAAAACAAATAA    ????????????????????
x2      0       ref2    2       30      21M     *       0       0       GGTTTTATAAAACAAATAATT   ?????????????????????
x3      0       ref2    6       30      9M4I13M *       0       0       TTATAAAACAAATAATTAAGTCTACA      ??????????????????????????
```

This feature is only available for indexed BAM, so this PR adds `indexed?` predicate to check the existence of index.

### DIFF
--- a/src/cljam/tools/cli.clj
+++ b/src/cljam/tools/cli.clj
@@ -44,12 +44,13 @@
   [[nil "--header" "Include header"]
    ["-f" "--format FORMAT" "Input file format <auto|sam|bam>"
     :default "auto"]
+   ["-r" "--region REGION" "Only print in region (e.g. chr6:1000-2000)"]
    ["-h" "--help" "Print help"]])
 
 (defn- view-usage [options-summary]
   (->> ["Extract/print all or sub alignments in SAM or BAM format."
         ""
-        "Usage: cljam view [--header] [-f FORMAT] <in.bam|sam>"
+        "Usage: cljam view [--header] [-f FORMAT] [-r REGION] <in.bam|sam>"
         ""
         "Options:"
         options-summary]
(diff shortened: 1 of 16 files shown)


src/cljam/tools/cli.clj
```diff
@@ -68,7 +71,13 @@
                                  "bam"  (sam/bam-reader f))]
         (when (:header options)
           (println (sam-util/stringify-header (sam/read-header r))))
-        (doseq [aln (sam/read-alignments r)]
+        (doseq [aln (if (:region options)
```
alumi: Why not adding `indexed?` to protocol `IAlignmentReader` or `IReader`?

totakke: Your question is reasonable. I hesitated over adding it to such protocols because they are also used for non-indexed readers like `SAMReader`. There are the following solutions:

* Not change - `indexed?` is internal function for the present.
* Add it to such protocols. Non-indexed readers always return `false`.
* Add a new protocol `IIndexedReader`.

I want to hear your opinion. Thanks.

alumi: Considering that we would have `CRAMReader` or DB or some other serialization formats in the future and that `SAMReader` already accepts `read-alignments` with 2 args, I think the 2nd option fits better to keep the code simple.
> Add it to such protocols. Non-indexed readers always return false.

As we want to know if a reader has an index or not,  a protocol to show a format is indexable is not so important.

totakke: OK, but I just realized that `indexed?` cannot be added to both `IAlignmentReader` and `ISequenceReader` because of name conflict. So it should be added to `IReader` and defined as a function checking the protocol can be random accessed.

totakke:
I added `indexed?` to `IReader` protocol. Please check it.

alumi:
LGTM! Thanks 👍

totakke:
Thank you for the advice.

PR #115, opened 2017-10-20
Don't refer clojure.core/indexed? in cljam.io.

alumi:
#### Summary

Since [`clojure.core/indexed?` will be added in Clojure 1.9](https://github.com/clojure/clojure/blob/a8f1c6436a8bfe181b0a00cf9e44845dcbbb63ee/src/clj/clojure/core.clj#L6199-L6202), this PR excludes it from `cljam.io` module to suppress warnings.

#### Tests

- `lein test :all` 🆗


totakke:
Sorry, I didn't notice that. Thanks.

PR #116, opened 2017-10-23
Move coverage phase to build stages

totakke:
The coverage phase accounts for about 2/3 of each build time on CI, and superfluous coverages are currently measured. By moving coverage phase to build stages, the coverage phase will be executed just once and a total of build time will be shortened.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,15 +14,16 @@ env:
     - VERSION='1.8'
     - VERSION='1.9'
 script: lein with-profile dev,$VERSION test
-after_success:
-  - CLOVERAGE_VERSION=1.0.9 lein cloverage -e '^cljam.main' --codecov
-  - bash <(curl -s https://codecov.io/bash) -f target/coverage/codecov.json
 jobs:
   include:
+    - stage: coverage
+      jdk: oraclejdk8
+      script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,15 +14,16 @@ env:
...
   include:
+    - stage: coverage
+      jdk: oraclejdk8
+      script:
+        - CLOVERAGE_VERSION=1.0.9 lein cloverage -e '^cljam.main' --codecov
+        - bash <(curl -s https://codecov.io/bash) -f target/coverage/codecov.json
     - stage: deploy
       jdk: oraclejdk8
       script: skip
-      after_success: echo skip
```
alumi: Is it because of an upgrade of Travis CI that this line's no longer necessary?
Seems working fine now, but I wonder why we introduced it.

alumi:
Merged. Thank you! 👍

.travis.yml
totakke: I added this line to avoid that coverage is executed on deploy stage because top-level `after_success` is also enabled on build stages. Now coverage stage doesn't use `after_success`, so this line isn't needed anymore.

totakke:
Thanks.

.travis.yml
alumi: I get it now, thanks! 😄

PR #117, opened 2017-10-25
Add FASTA-TwoBit converter

totakke:
Adds FASTA-TwoBit conversion, and test cases also.

As known issue, it is hard to convert a large reference file such as `hg38.fa` because of memory consumption. The reason is that `cljam.io.fasta.reader/sequential-read-string` is non-lazy and `cljam.io.twobit.reader/read-all-sequences` realizes some sequences by chunking. I plan on fixing them on another branch.


alumi:
LGTM 👍 
>  I plan on fixing them on another branch.

And thank you for working on the memory issue!

PR #119, opened 2017-11-02
Fix a bug that 2bit reader incorrectly reads small n

totakke:
Current 2bit reader cannot reads small `n` correctly even if `:mask? true` is specified.

For example, when reading the following 2bit file,

```
>ref
nnNN
```

`n` are replaced with `N`.

```clojure
(with-open [rdr (cseq/reader "path/to/nnNN.2bit")]
  (doall (cseq/read-all-sequences rdr {:mask? true})))
;;=> ({:name "ref", :sequence "NNNN"})
```

This patch fixes that issue.

```clojure
(with-open [rdr (cseq/reader "path/to/nnNN.2bit")]
  (doall (cseq/read-all-sequences rdr {:mask? true})))
;;=> ({:name "ref", :sequence "nnNN"})
```


alumi:
Merged. Thank you 👍

PR #120, opened 2017-11-09
Improve performance of FASTA random reading

alumi:
#### Summary
Improve performance of FASTA random reading by **10x** using MappedByteBuffer and CharBuffer.

#### Changes
Current implementation of `read-sequence` for FASTA is slow because of multiple String generation.
byte array => String => String => String

This PR uses mmap and CharBuffer to generate String once.
MappedByteBuffer => CharBuffer => String

##### benchmarking code
```clojure
(are [reg opts] (c/quick-bench (with-open [rdr (cseq/reader "/path/to/hg38.fa")]
                                 (dorun (cseq/read-sequence rdr reg opts))))
  {:chr "chr1" :start 1 :end 248956422} {:mask? true}
  {:chr "chr1" :start 1 :end 248956422} {})
```

##### master
```
[{:chr "chr1", :start 1, :end 248956422} {:mask? true}]
time: 20.807584 sec, sd: 23.176079 ms
[{:chr "chr1", :start 1, :end 248956422} {}]
time: 22.297717 sec, sd: 30.394828 ms
```

##### this branch
```
[{:chr "chr1", :start 1, :end 248956422} {:mask? true}]
time: 1.994279 sec, sd: 9.574159 µs
[{:chr "chr1", :start 1, :end 248956422} {}]
time: 2.007051 sec, sd: 312.275365 µs
```

#### Tests

- `lein test :all` 🆗

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: clojure
-lein: 2.8.0
+lein: 2.8.1
 cache:
   directories:
     - $HOME/.m2
--- a/project.clj
+++ b/project.clj
@@ -30,7 +30,7 @@
                    :global-vars {*warn-on-reflection* true}}
              :1.7 {:dependencies [[org.clojure/clojure "1.7.0"]]}
              :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
-             :1.9 {:dependencies [[org.clojure/clojure "1.9.0-beta2"]]}
+             :1.9 {:dependencies [[org.clojure/clojure "1.9.0-RC1"]]}
              :uberjar {:dependencies [[org.clojure/clojure "1.8.0"]
                                       [org.apache.logging.log4j/log4j-api "2.9.1"]
                                       [org.apache.logging.log4j/log4j-core "2.9.1"]]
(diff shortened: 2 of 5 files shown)


alumi:
Thank you! 😃

totakke:
Although not related to this change, I felt the way of handling external region is somewhat strange.

When the supplied region is partially over the sequence, missing bases are filled with `N`.

```clojure
(with-open [rdr (reader "test-resources/fasta/test.fa")]
  (read-sequence rdr {:chr "ref" :start 30 :end 60}))
;;=> AGGCAGTCAGCGCCATNNNNNNNNNNNNNNN
```

But when the region is not included in sequence, `read-sequence` returns `nil`.

```clojure
(with-open [rdr (reader "test-resources/fasta/test.fa")]
   (read-sequence rdr {:chr "ref" :start 50 :end 60}))
;;=> nil
```

This behavior seems inconsistent. What do you think about this?

alumi:
You're absolutely right!
I think we don't need padding because no other readers do pad anything.
I'll create a PR for this so please check it later.

Thank you.

PR #123, opened 2017-12-06
Fix parsing SAM header values containing ':'.

alumi:
#### Summary
Don't discard substring after second `:` when parsing SAM header.

#### Problem
Complex SAM headers can be modified unexpectedly.

##### input sam
```
@SQ	SN:A:B;100-200;++;chr1:100-200	LN:400
```

##### output sam
```
@SQ	SN:A	LN:400
```

#### Cause
```clojure
(clojure.string/split "SN:A:B;100-200;++;chr1:100-200" #":")
; => ["SN" "A" "B;100-200;++;chr1" "100-200"]
```
`clojure.string/split` with 2 args splits given string at all ocurrence of the regex.
#### Changes
Pass `2` as 3rd arg (`limit`) of `clojure.string/split`.

#### Affects
SAM/BAM I/O

#### Tests

- `lein test :all` 🆗

### DIFF
--- a/src/cljam/io/sam/util.clj
+++ b/src/cljam/io/sam/util.clj
@@ -16,7 +16,7 @@
   (into
    {}
    (map (fn [kv]
-          (let [[k v] (cstr/split kv #":")]
+          (let [[k v] (cstr/split kv #":" 2)]
             [(keyword k)
              (case k
                "LN" (Integer/parseInt v)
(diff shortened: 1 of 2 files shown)


totakke:
Thank you 😄

alumi:
Thank you for reviewing! 😸

PR #124, opened 2017-12-06
Add some extensions for FASTA format.

alumi:
Add some file extensions of FASTA format listed in https://en.wikipedia.org/wiki/FASTA_format#File_extension and http://fileformats.archiveteam.org/wiki/FASTA_and_FASTQ.


alumi:
@totakke Thank you for pointing that. I forgot to check the index.

totakke:
LGTM. Thanks.

PR #125, opened 2017-12-15
Transducers and Reducing Functions for writing SAM/BAM and BAI.

alumi:
#### Summary
This PR adds transducers and reducing functions for SAM/BAM and BAI writers to make it easy to reuse and to combine.

I think it's worth implementing other `io` and `algo` functions as transducers / reducing functions as future works.

#### Tests

- `lein test :all` 🆗

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -17,7 +17,7 @@
                                   [criterium "0.4.4"]
                                   [net.totakke/libra "0.1.1"]
                                   [org.tcrawley/dynapath "1.0.0"]]
-                   :plugins [[lein-binplus "0.6.2" :exclusions [org.clojure/clojure]]
+                   :plugins [[lein-binplus "0.6.3" :exclusions [org.clojure/clojure]]
                              [lein-codox "0.10.3"]
                              [lein-marginalia "0.9.1" :exclusions [org.clojure/clojure]]
                              [lein-cloverage "1.0.10" :exclusions [org.clojure/clojure org.tcrawley/dynapath]]
--- a/project.clj
+++ b/project.clj
@@ -15,13 +15,13 @@
   :profiles {:dev {:dependencies [[org.clojure/clojure "1.8.0"]
                                   [cavia "0.4.3"]
                                   [criterium "0.4.4"]
-                                  [net.totakke/libra "0.1.0"]
+                                  [net.totakke/libra "0.1.1"]
                                   [org.tcrawley/dynapath "1.0.0"]]
                    :plugins [[lein-binplus "0.6.2" :exclusions [org.clojure/clojure]]
                              [lein-codox "0.10.3"]
                              [lein-marginalia "0.9.1" :exclusions [org.clojure/clojure]]
                              [lein-cloverage "1.0.10" :exclusions [org.clojure/clojure org.tcrawley/dynapath]]
-                             [net.totakke/lein-libra "0.1.0"]]
+                             [net.totakke/lein-libra "0.1.1"]]
                    :test-selectors {:default #(not-any? % [:slow :remote])
                                     :slow :slow ; Slow tests with local resources
                                     :remote :remote ; Tests with remote resources
--- a/project.clj
+++ b/project.clj
@@ -32,8 +32,8 @@
              :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
              :1.9 {:dependencies [[org.clojure/clojure "1.9.0-RC1"]]}
              :uberjar {:dependencies [[org.clojure/clojure "1.8.0"]
-                                      [org.apache.logging.log4j/log4j-api "2.9.1"]
-                                      [org.apache.logging.log4j/log4j-core "2.9.1"]]
+                                      [org.apache.logging.log4j/log4j-api "2.10.0"]
+                                      [org.apache.logging.log4j/log4j-core "2.10.0"]]
                        :resource-paths ["bin-resources"]
                        :main cljam.tools.main
                        :jvm-opts ["-Dclojure.compiler.direct-linking=true"]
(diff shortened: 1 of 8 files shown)


totakke:
I agree with you 😄  Providing transducers seems very convenient, but why do you add `-rf` or `-xf` functions? Sequence functions in `clojure.core`, such as `map`, return a transducer when no data is provided. I think such style is more consistent and conventional. Are there any reason for separating the functions?

alumi:
- new `write-alignments-rf` returns a reducing function which writes headers and refs
- new `write-alignments-xf` returns a transducer and writes headers and refs
- existing `write-alignments` does the reduction in it and writes alignments only

It is possible to add a 3-arity version of `write-alignments-xf` which does `(transduce (write-alignments-xf writer header) (fn [& _]) coll)` in it, but it's not the same behavior as existing `write-alignments` because `write-alignments-xf` writes headers and refs, too.

It is also possible to add the 2-arity version of `write-alignments` which returns a transducer without writing headers and refs.

What do you think about how should we merge these functions?
Thank you

totakke:
I can't understand the reason why their functions need to write headers/refs. `write-header/refs` and `write-alignments` are separated processes from the first, so we can assume header/refs are written before writing alignments. If you want to strictly limit the order, there is another way that `BAMWriter` constructor receives header/refs like `VCFWriter`.

alumi:
Thank you for the comment!
I'm really unconfident how we should manage headers & refs 😟 
Please let me clarify problems about introducing transducers to cljam.

---

What we want to do is providing composable and reusable functions for manipulating alignment data.
For example, 'simultaneously count all primary reads, split the file into mapped/unmapped and collect the name of unmapped reads' may be written as follows:

```clojure

(defn the-task [sam-writer1 sam-writer2 header]
  ((filter (comp sam-util/primary? :flag))
   (juxt-rf
    count-rf
    ((remove (comp :unmapped sam-util/decode-flags :flag))
     (sam/write-alignments-rf sam-writer1 header))
    ((filter (comp :unmapped sam-util/decode-flags :flag))
     (juxt-rf
       (sam/write-alignments-rf sam-writer2 header)
       ((map :qname) into-vec))))))

(transduce
 identity
 (the-task sam-writer1 sam-writer2 (sam/read-header sam-reader))
 (sam/read-alignments sam-reader))
```

where some utility functions pre-defined:

```clojure
(defn juxt-rf [& rfs]
  (fn
    ([] (mapv #(%) rfs))
    ([result] (mapv #(%1 (unreduced %2)) rfs result))
    ([result input]
     (let [ret (mapv #(if (reduced? %2) %2 (%1 %2 input)) rfs result)]
       (if (every? reduced? ret)
         (reduced (mapv unreduced ret))
         ret)))))

(defn count-rf
  ([] 0)
  ([result] result)
  ([result _] (inc result)))

(defn into-vec
  ([] (transient []))
  ([result] (persistent! result))
  ([result input] (conj! result input)))
```

These processes can grow much more complex and I think it's not handy to manage such headers & refs outside of the main task.

```clojure
(let [header (sam/read-header sam-reader)]
  (doseq [w [sam-writer-1 sam-writer-2 ...]]
    (sam/write-header w header)
    (sam/write-refs w header)))
```

Writing header & refs in the `BAMWriter` constructors would be better, but it breaks the compatibility.
I'll open another PR #126 for that to check how it feels like.
So please take a look at it, I'd appreciate it.

---

However, there's another problem with transforming headers.
If we write headers & refs in the constructors, we have to
1. change/wrap `cljam.algo.sorter/sort!` and `cljam.algo.normalize/normalize` so that the functions take filenames instead of reader & writer
2. or manage how to transform the header when constructing the writer outside of the functions

We can choose the first option for simple file I/O but when combined with transducers, we can't.
Algorithms like `convert`, `dedupe`, `level`, `depth` and `pileup` can be rewritten as transducers and reducing functions.
It would be nice if we could also have `sort-by` and `normalize` transducer which can be used like:

```clojure
(def sort-and-normalize
 (comp
  (algo/sort-by :coordinate)
  (algo/normalize)))
```

But we have to write another transformation for the header:

```clojure
(def sort-and-normalize-header
  (comp
   algo/normalized-header
   (partial algo/sorted-header :coordinate)))
```

and then create a reducing function like:

```clojure
(defn wrapped-task [sam-writer1 sam-writer2]
  (sort-and-normalize (the-task sam-writer1 sam-writer2)))

(def wrapped-task-header sort-and-normalize-header)

(with-open [r (sam/reader in)]
  (let [header (wrapped-task-header (sam/read-header r))]
    (with-open [w1 (sam/writer out1 header)
                w2 (sam/writer out2 header)]
      (transduce
       identity
       (wrapped-task w1 w2)
       (sam/read-alignments r)))))
```

or

```clojure
(defn wrapped-task-and-writers [out1 out2 header]
  (let [hdr (sort-and-normalize-header header)
        w1 (sam/writer out1 hdr)
        w2 (sam/writer out2 hdr)]
    [w1 w2 (sort-and-normalize (the-task sam-writer1 sam-writer2))]))

(with-open [r (sam/reader in)]
  (let [[w1 w2 task] (wrapped-task-and-writers out1 out2 (sam/read-header r))]
    (with-open [w1 w1
                w2 w2]
      (transduce identity task (sam/read-alignments r)))))
```

instead of

```clojure
(defn wrapped-task [sam-writer1 sam-writer2 header]
  (let [hdr (sort-and-normalize-header header)]
    (sort-and-normalize (the-task sam-writer1 sam-writer2 hdr))))

(with-open [r (sam/reader in)
            w1 (sam/writer out1)
            w2 (sam/writer out2)]
  (transduce
   identity
   (wrapped-task w1 w2 (sam/read-header r))
   (sam/read-alignments r)))
```

To me, it's difficult to decide which is better and I currently have no better ideas to deal with this...

Sorry for writing such a long comment.

PR #128, opened 2017-12-21
Add many util functions for SAM/BAM files & split cljam.io.sam.util.

alumi:
#### Summary
This PR adds utility functions for SAM/BAM files.
- parsers for optional fields
    - According to [SAM spec](https://github.com/samtools/hts-specs/blob/master/SAMv1.pdf), [optional fields spec](https://github.com/samtools/hts-specs/blob/master/SAMtags.pdf) and [BWA](http://bio-bwa.sourceforge.net/bwa.shtml#tbl).
- reverse-complement sequence
- predicates for flags

Because many functions are placed in `cljam.io.sam.util`,
I split the namespace into some sub-namespaces to keep them organized.

#### Tests

- `lein test :all` 🆗


src/cljam/io/sam/util/sequence.clj
```diff
@@ -0,0 +1,101 @@
...
+(def ^:private revcomp-table
+  (let [ba (byte-array 128)]
+    (dotimes [i 128] (aset-byte ba i (byte \N)))
+    (doseq [s ["AT" "GC" "NN"]
+            [a b] [(cstr/upper-case s) (cstr/lower-case s)]]
+      (aset-byte ba (int a) (byte b))
+      (aset-byte ba (int b) (byte a)))
+    ba))
+
+(defn reverse-complement
```
totakke: `reverse-complement` is a general process in genomic analysis, not just for SAM. I recommend moving it to top-level utils (`cljam.util.*`).

Besides, I personally feel `revcomp` is better name because it is shorter and enough understandable.

src/cljam/algo/sorter.clj
```diff
@@ -174,25 +165,21 @@
...
+  (sort! rdr wtr (into (or option {}) {:mode header/order-queryname})))
 
 (defn sorted-by?
   "Returns true if the sam is sorted, false if not. It is detected by
   `@HD SO:***` tag in the header."
   [rdr]
-  (let [so (:SO (:HD (sam/read-header rdr)))]
-    (or (= so (name order-queryname))
-        (= so (name order-coordinate)))))
+  (header/sorted-by? (sam/read-header rdr)))
```
totakke: Though it is not related to this PR, why was this function named `sorted-by?` ...? `-by` seems unnecessary.

src/cljam/io/sam/util/sequence.clj
alumi: Thank you for the advice.
I renamed it to `cljam.util.sequence/revcomp`.

src/cljam/algo/sorter.clj
alumi: I agree. Renamed it to `sorted?`.

alumi:
Thank you for the reviewing! 😸

totakke:
Thanks :)

PR #130, opened 2018-04-04
Fix a bug in merging bed regions

alumi:
#### Summary
Fix a bug in `cljam.io.bed/merge-fields` that adjacent regions are not properly merged.

#### Changes
Use a function in `cljam.util.region` to merge + upgrade dependencies

#### Tests

- `lein test :all` 🆗


totakke:
Thanks!

PR #131, opened 2018-05-29
Add some BED manipulation APIs

athos:
This PR adds the following new APIs to `cljam.io.bed`:

- `intersect-fields`
- `subtract-fields`
- `complement-fields`

Each implementation reflects the behavior of `bedtools`' corresponding subcommand. See  the links below for details:

- intersect: http://bedtools.readthedocs.io/en/latest/content/tools/intersect.html
- subtract: http://bedtools.readthedocs.io/en/latest/content/tools/subtract.html
- complement: http://bedtools.readthedocs.io/en/latest/content/tools/complement.html

There is one thing to be confirmed.

`complement-fields` now throws an exception if the map passed as the first argument (`chrom-lengths`) lacks the entry for some chromosome appeared in the sequence of BED fields. However, the exception will be deferred until the resulting sequence is realized because `complement-fields` returns a lazy sequence and the body of `complement-fields` is wrapped with `lazy-seq`.

The design of throwing exceptions from lazy sequences doesn't look very good to me. Is it better to scan the input sequences and check if every chromosome appeared in them can be found in `chrom-lengths` before processing the sequences? Any suggestions?

### DIFF
--- a/src/cljam/io/bed.clj
+++ b/src/cljam/io/bed.clj
@@ -208,6 +208,86 @@
    0
    (sort-fields xs)))
 
+(defn- fields-<
+  "Compare two BED fields on the basis of :chr and :end fields."
+  [x y]
+  (< (compare [(chr/chromosome-order-key (:chr x)) (:end x)]
+              [(chr/chromosome-order-key (:chr y)) (:end y)])
+     0))
+
+(defn intersect-fields
+  "Returns a lazy sequence that is the intersection of the two BED sequences.
+
+  The input sequences will first be sorted with sort-fields, which may cause
+  an extensive memory use for ones with a large number of elements."
+  [xs ys]
+  (letfn [(intersect [xs ys]
+            (lazy-seq
+             (if-let [x (first xs)]
+               (if-let [y (first ys)]
+                 (cond->> (if (fields-< x y)
+                            (intersect (next xs) ys)
+                            (intersect xs (next ys)))
+                   (region/overlapped-regions? x y)
+                   (cons (-> x
+                             (update :start max (:start y))
+                             (update :end min (:end y)))))
+                 [])
+               [])))]
+    (intersect (sort-fields xs) (sort-fields ys))))
+
+(defn subtract-fields
+  "Returns a lazy sequence that is the result of subtracting the BED fields
+  in the sequence ys from the sequence xs.
+
+  The input sequences will first be sorted with sort-fields, which may cause
+  an extensive memory use for ones with a large number of elements."
+  [xs ys]
+  (letfn [(subtract [xs ys]
+            (lazy-seq
+             (if-let [x (first xs)]
+               (if-let [y (first ys)]
+                 (let [[r1 r2] (region/subtract-region x y)]
+                   (if r2
+                     (cons r1 (subtract (cons r2 (next xs)) (next ys)))
+                     (if r1
+                       (if (fields-< r1 y)
+                         (cons r1 (subtract (next xs) ys))
+                         (subtract (cons r1 (next xs)) (next ys)))
+                       (if (fields-< x y)
+                         (subtract (next xs) ys)
+                         (subtract xs (next ys))))))
+                 xs)
+               [])))]
+    (subtract (sort-fields xs) (sort-fields ys))))
+
+(defn complement-fields
+  "Takes a map of chromosome name to its length and a sequence of BED fields,
+  and returns a lazy sequence that is the complement of the sequence.
+
+  The input sequence will first be sorted with sort-fields, which may cause
+  an extensive memory use for ones with a large number of elements."
+  [chrom-lengths xs]
+  (let [chrs (sort-by chr/chromosome-order-key (keys chrom-lengths))]
+    (letfn [(complement [xs chrs pos]
+              (lazy-seq
+               (if-let [chr (first chrs)]
+                 (let [len (get chrom-lengths chr)
+                       x (first xs)]
+                   (if (and x (= (:chr x) chr))
+                     (cond->> (complement (next xs) chrs (:end x))
+                       (< pos (:start x))
+                       (cons {:chr chr :start pos :end (:start x)}))
+                     (cond->> (complement xs (next chrs) 0)
+                       (< pos len)
+                       (cons {:chr chr :start pos :end len}))))
+                 (when (seq xs)
+                   (let [msg (str "Length of chromosome "
+                                  (:chr (first xs))
+                                  " not specified")]
+                     (throw (IllegalArgumentException. msg)))))))]
+      (complement (sort-fields xs) chrs 0))))
+
 (defn write-raw-fields
   "Write sequence of BED fields to writer without converting :start and :thick-start values."
   [^BEDWriter wtr xs]
--- a/src/cljam/io/bed.clj
+++ b/src/cljam/io/bed.clj
@@ -275,10 +275,10 @@
                  (let [len (get chrom-lengths chr)
                        x (first xs)]
                    (if (and x (= (:chr x) chr))
-                     (cond->> (complement (next xs) chrs (:end x))
+                     (cond->> (complement (next xs) chrs (inc (:end x)))
                        (< pos (:start x))
-                       (cons {:chr chr :start pos :end (:start x)}))
-                     (cond->> (complement xs (next chrs) 0)
+                       (cons {:chr chr :start pos :end (dec (:start x))}))
+                     (cond->> (complement xs (next chrs) 1)
                        (< pos len)
                        (cons {:chr chr :start pos :end len}))))
                  (when (seq xs)
@@ -286,7 +286,7 @@
                                   (:chr (first xs))
                                   " not specified")]
                      (throw (IllegalArgumentException. msg)))))))]
-      (complement (sort-fields xs) chrs 0))))
+      (complement (sort-fields xs) chrs 1))))
 
 (defn write-raw-fields
   "Write sequence of BED fields to writer without converting :start and :thick-start values."
--- a/src/cljam/io/bed.clj
+++ b/src/cljam/io/bed.clj
@@ -219,7 +219,11 @@
   "Returns a lazy sequence that is the intersection of the two BED sequences.
 
   The input sequences will first be sorted with sort-fields, which may cause
-  an extensive memory use for ones with a large number of elements."
+  an extensive memory use for ones with a large number of elements.
+  Note also that this function assumes the input sequences contain only valid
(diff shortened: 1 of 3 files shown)


src/cljam/io/bed.clj
```diff
@@ -208,6 +208,86 @@
...
+    (subtract (sort-fields xs) (sort-fields ys))))
+
+(defn complement-fields
+  "Takes a map of chromosome name to its length and a sequence of BED fields,
+  and returns a lazy sequence that is the complement of the sequence.
+
+  The input sequence will first be sorted with sort-fields, which may cause
+  an extensive memory use for ones with a large number of elements."
+  [chrom-lengths xs]
+  (let [chrs (sort-by chr/chromosome-order-key (keys chrom-lengths))]
```
alumi: Lengths of chromosomes can be obtained by [cljam.io.sam/read-refs](https://github.com/chrovis/cljam/blob/626320bdb1d6808e89e8ca494db6de3c2d2101a6/test/cljam/test_common.clj#L280-L303) or [cljam.io.sequence/read-indices](https://github.com/chrovis/cljam/blob/626320bdb1d6808e89e8ca494db6de3c2d2101a6/test/cljam/io/sequence_test.clj#L31-L41), but to make up a map we need to write something like `(into {} (map (juxt :name :len)) ~~~)` which I think is a bit redundant.

How about changing `(keys chrom-lengths)` to `(map first chrom-lengths)` or accepting a sequence of maps containing `:name` and `:len`?

src/cljam/io/bed.clj
```diff
@@ -208,6 +208,86 @@
...
+                 (let [[r1 r2] (region/subtract-region x y)]
+                   (if r2
+                     (cons r1 (subtract (cons r2 (next xs)) (next ys)))
+                     (if r1
+                       (if (fields-< r1 y)
+                         (cons r1 (subtract (next xs) ys))
+                         (subtract (cons r1 (next xs)) (next ys)))
+                       (if (fields-< x y)
+                         (subtract (next xs) ys)
+                         (subtract xs (next ys))))))
```
alumi: Please check the following case.

```clojure
(cljam.io.bed/subtract-fields
  [{:chr "chr1" :start 1 :end 10} {:chr "chr1" :start 1 :end 11}]
  [{:chr "chr1" :start 1 :end 10} {:chr "chr1" :start 9 :end 11}])
=> ({:chr "chr1", :start 1, :end 8} {:chr "chr1", :start 1, :end 8})
```

src/cljam/io/bed.clj
```diff
@@ -208,6 +208,86 @@
...
+                 (cond->> (if (fields-< x y)
+                            (intersect (next xs) ys)
+                            (intersect xs (next ys)))
+                   (region/overlapped-regions? x y)
+                   (cons (-> x
+                             (update :start max (:start y))
+                             (update :end min (:end y)))))
+                 [])
+               [])))]
+    (intersect (sort-fields xs) (sort-fields ys))))
```
alumi: Please check the following case.

```clojure
(cljam.io.bed/intersect-fields
  [{:chr "chr1" :start 1 :end 10} {:chr "chr1" :start 1 :end 11}]
  [{:chr "chr1" :start 1 :end 10} {:chr "chr1" :start 9 :end 11}])
=> ({:chr "chr1", :start 1, :end 10} {:chr "chr1", :start 9, :end 10} {:chr "chr1", :start 9, :end 11})
```

alumi:
> The design of throwing exceptions from lazy sequences doesn't look very good to me. Is it better to scan the input sequences and check if every chromosome appeared in them can be found in chrom-lengths before processing the sequences? Any suggestions?

Though there're already some fns throwing exceptions from lazy sequences in cljam, I think pre-checking is better in this particular case because all arguments are realized.

athos:
> Looks like working fine with common cases but not for some edge cases because BED file format permits overlapping regions.

Thank you for pointing it out! I didn't know that. I'll fix it soon.

> I think pre-checking is better in this particular case because all arguments are realized.

I see. I'll give it a try in that way.

src/cljam/io/bed.clj
athos: So, considering cooperation with those other functions, it would be more useful to accept such a data structure, right?

OK, I'm going to change the function's signature to accept a sequence of maps like `{:name ..., :len ...}`.

totakke: I also feel `({:name :len})` is more consistent.

test/cljam/io/bed_test.clj
```diff
@@ -172,6 +172,89 @@
...
+    [{:chr "chr1" :start 1 :end 1000} {:chr "chr2" :start 1 :end 800}]
+
+    [{:chr "chr1" :start 1 :end 300}]
+    [{:chr "chr1" :start 301 :end 1000} {:chr "chr2" :start 1 :end 800}]
+
+    [{:chr "chr1" :start 1 :end 300} {:chr "chr1" :start 900 :end 1000} {:chr "chr2" :start 1 :end 300}]
+    [{:chr "chr1" :start 301 :end 899} {:chr "chr2" :start 301 :end 800}])
+
+  (is (thrown? IllegalArgumentException
+               (doall (bed/complement-fields {"chr1" 1000} [{:chr "chr2" :start 1 :end 100}])))))
```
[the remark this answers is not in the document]
athos: Ah, exactly. The prechecking discussed above will remove this explicit realization, though.

athos:
Thank you for your review.

> I found that result is strange when illegal regions are supplied, such as :start > :end.

Exactly, these fns have a precondition that the input sequences contain only valid regions, so once the condition is violated they can probably produce garbage.

I'm not sure it's reasonable to check region's validity in all these fns. If doing so, and then if you want to use some of these fns combined (say, you want the complement of the intersection of BED sequences `xs` and `ys`), the validity check will be run more than once unnecessarily.

I would rather propose adding a new API to help users to check the validity of a BED sequence on their own, something like the sequence version of `util.region/valid-region?`.

totakke:
If not checking, I think you don't have to add a new validator function. It can be easily written by `every? util.region/valid-region?`. Instead, you should add a statement to the docstring that behavior is not supported when illegal regions are supplied.

athos:
I just fixed the issues other than the one about overlapping regions and the one about the region validity.

As for the region validity, I'll update the docstrings for these fns to describe their limitation of failing to handle sequences containing invalid regions, as @totakke suggested immediately above.

As for overlapping regions, I'm still wondering how I should resolve the issue. Is it a good approach to first apply `merge-fields` to the input sequences instead of just doing `sort-fields`? I have no idea what the expected result should be for overlapping cases.

alumi:
> As for overlapping regions, I'm still wondering how I should resolve the issue. Is it a good approach to first apply merge-fields to the input sequences instead of just doing sort-fields? I have no idea what the expected result should be for overlapping cases.

It's just my opinion but, I think merging `ys` is reasonable.

`bedtools subtract` acts like that:

```console
$ bedtools --version
bedtools v2.27.1
$ bedtools subtract -a <(echo -e 'chr1\t0\t10\tA\nchr1\t0\t11\tB') -b <(echo -e 'chr1\t0\t10\tC\nchr1\t8\t11\tD')
$ bedtools subtract -a <(echo -e 'chr1\t0\t10\tA\nchr1\t0\t11\tB') -b <(echo -e 'chr1\t1\t10\tC\nchr1\t8\t11\tD')
chr1	0	1	A
chr1	0	1	B
```

`bedtools intersect` produces results that are unintuitive to me:

```console
$ bedtools intersect -a <(echo -e 'chr1\t0\t10\tA\nchr1\t0\t11\tB') -b <(echo -e 'chr1\t0\t10\tC\nchr1\t8\t11\tD')
chr1	0	10	A
chr1	8	10	A
chr1	0	10	B
chr1	8	11	B
$ bedtools intersect -a <(echo -e 'chr1\t0\t10\tA\nchr1\t0\t11\tB') -b <(echo -e 'chr1\t0\t10\tC') -b <(echo -e 'chr1\t8\t11\tD')
chr1	0	10	A
chr1	8	10	A
chr1	0	10	B
chr1	8	11	B
```

merging `ys` gives followings:

```console
$ bedtools intersect -a <(echo -e 'chr1\t0\t10\tA\nchr1\t0\t11\tB') -b <(bedtools merge -i <(echo -e 'chr1\t0\t10\tC\nchr1\t8\t11\tD'))
chr1	0	10	A
chr1	0	11	B
```

and I think this is more consistent with:

```console
$ bedtools subtract -a <(echo -e 'chr1\t0\t10\tA\nchr1\t0\t11\tB') -b <(bedtools subtract -a <(echo -e 'chr1\t0\t10\tA\nchr1\t0\t11\tB') -b <(echo -e 'chr1\t0\t10\tC\nchr1\t8\t11\tD'))
chr1	0	10	A
chr1	0	11	B
$ bedtools subtract -a <(echo -e 'chr1\t0\t10\tA\nchr1\t0\t11\tB') -b <(bedtools complement -i <(echo -e 'chr1\t0\t10\tC\nchr1\t8\t11\tD') -g <(echo -e 'chr1\t20'))
chr1	0	10	A
chr1	0	11	B
```

totakke:
That may be because `bedtools intersect` accepts unsorted regions. `bedtools` clearly distinguishes `sort` and other processes.

If sorting is included in a process, I feel merging input regions is intuitive and reasonable.

athos:
Thank you for the detailed explanation!

The results from merged inputs look reasonable to me as well. So, I'm going to change the fns so that they apply `merge-fields` instead of `sort-fields` to their second argument.

athos:
Done.

Along with the fix, I modified `fields-<` to tweak which sequence should be taken next when the current head elements of them have the same `:end`.

src/cljam/io/bed.clj
```diff
@@ -208,6 +208,98 @@
...
+             (if-let [x (first xs)]
+               (if-let [y (first ys)]
+                 (let [[r1 r2] (region/subtract-region x y)]
+                   (if r2
+                     (cons r1 (subtract (cons r2 (next xs)) (next ys)))
+                     (if r1
+                       (if (fields-<= r1 y)
+                         (cons r1 (subtract (next xs) ys))
+                         (subtract (cons r1 (next xs)) (next ys)))
+                       (if (fields-<= x y)
```
alumi: Seems to be always true because `x` is totally covered by `y` 🤔

athos: Ah, right. Good catch! 🙏

totakke:
Thank you for good enhancement.

athos:
Thank you for a lot of advice and comments! I couldn't do this alone 😂

PR #132, opened 2018-06-21
Support more source types

totakke:
Current `cljam.io` basically supports only a string file path as input/output.
This PR adds support for more source types such as `java.io.File` or `java.net.URL`.

```clojure
(require '[cljam.io.sam :as sam])

(with-open [r (sam/reader (java.io.File. "path/to/test.sam"))]
  (first (sam/read-alignments r)))

(with-open [r (sam/reader (java.net.URL. "file:/path/to/test.sam"))]
  (first (sam/read-alignments r)))
```

http source also can be used except fasta/2bit readers.

```clojure
(with-open [r (sam/reader (java.net.URL. "http://example.com/path/to/test.sam"))]
  (first (sam/read-alignments r)))
```

Reader/writer keep a file location as `java.net.URL`, and protocol methods
`{reader,writer}-path` are renamed to `{reader,writer}-url`. This may be
breaking changes.

I confirmed `test :all` passed.


src/cljam/io/bam/core.clj
```diff
@@ -1,56 +1,57 @@
...
-    (bai/bam-index bai-path)
-    (throw (FileNotFoundException. (str "Could not find BAM Index file for " bam-path)))))
+  "Loads an index file (BAI) for the given BAM URL."
+  [bam-url]
+  (or (->> ["$1.bai" ".bai" "$1.BAI" ".BAI"]
+           (eduction
+            (comp
+             (map #(cstr/replace (str bam-url) #"(?i)(\.bam)$" %))
+             (map util/as-url)
+             (map #(try (bai/bam-index %) (catch FileNotFoundException _)))))
```
alumi: `map` => `keep` or `(first (eduction (comp (map A) (map B) (map C)) coll))` => `(some (comp A B C) coll)`

src/cljam/io/fasta/core.clj
```diff
@@ -13,33 +13,33 @@
...
-                           first)]
-      (fai/reader fai-path)
-      (throw (FileNotFoundException.
-              (str "Could not find FASTA Index file for " fasta-path))))))
+    (or (->> ["$1.fai" ".fai" "$1.FAI" ".FAI"]
+             (eduction
+              (comp
+               (map #(cstr/replace (str fasta-url) fasta-exts %))
+               (map util/as-url)
+               (map #(try (fai/reader %) (catch FileNotFoundException _)))))
```
alumi: `map` => `keep` or `(first (eduction (comp (map A) (map B) (map C)) coll))` => `(some (comp A B C) coll)`

totakke:
Thank you for pointing out my mistake. I fixed it.

alumi:
Renaming `{reader,writer}-path` are not so big changes and I'm not using these fns in other projects. I think we can merge this PR into `0.5.2-SNAPSHOT`. Is that OK with you @totakke?

totakke:
👌

PR #135, opened 2018-07-02
Support extra fields for VCF/BCF meta info

athos:
According to [VCF spec v4.3](https://github.com/samtools/hts-specs/blob/ce7f2a61fca5e9f906caa72128cade0065b8361a/VCFv4.3.tex#L106-L112), "For all of the structured lines (##INFO, ##FORMAT, ##FILTER, etc.), extra fields can be included after the default fields."

cljam now accepts any kind of fields, but only emits the required fields and a few relatively common optional ones. This PR makes it fully support extra fields, and enables to emit them as users want.


totakke:
My comment was also from the spec, so that I think support of `\"` and `\\` is necessary and enough. I leave you to separate PR or not.

athos:
OK, then I'll open another PR for that later.

athos:
Thank you for detailed comments!
I'm still not so familiar with the spec, so those comments on the spec details are very helpful 🙏 

I have a few things to be confirmed:
- Several (all?) kinds of meta info are supporting the `idx` field, so are special treatments for it necessary for all of them?
- I'm still not sure how the PEDIGREE meta info line should be formed. My understanding (from your comments and the spec) is that it only requires the `id` field and optionally takes other fields of any kind (as extra fields). Is it correct?

alumi:
> - Several (all?) kinds of meta info are supporting the idx field, so are special treatments for it necessary for all of them?

Yes. My understanding is that `IDX` is used to refer meta info from variant records in BCF: `contig` for CHROM, `FILTER` for FILTER, `INFO` for INFO and `FORMAT` for FORMAT fields.
- `contig`s have distinct `IDX`s.
    - https://github.com/samtools/htslib/blob/be22a2a1082f6e570718439b9ace2db17a609eae/vcf.c#L403-L446
- `FILTER`, `INFO` and `FORMAT` are sharing their `IDX`s.
    - https://github.com/samtools/htslib/blob/be22a2a1082f6e570718439b9ace2db17a609eae/vcf.c#L458-L497
- `ALT`s don't have `IDX`s because ALT fields are encoded as typed strings. 
    - https://github.com/samtools/htslib/blob/be22a2a1082f6e570718439b9ace2db17a609eae/vcf.c#L1285-L1292
- Other meta info like `SAMPLE` and `PEDIGREE` have no corresponding fields in variant data, thus they don't have `IDX`.

> - I'm still not sure how the PEDIGREE meta info line should be formed. My understanding (from your comments and the spec) is that it only requires the id field and optionally takes other fields of any kind (as extra fields). Is it correct?

I belive so, too.

athos:
Thanks a lot! I'll fix them soon.

athos:
Just fixed what @alumi pointed out.

Also I created [a new issue](https://github.com/chrovis/cljam/issues/136) for enhancing character escaping support for VCF string fields. For further development for that, watch that issue and the succeeding PR.

athos:
@totakke @alumi I'm working on #136 (character escaping enhancement). The branch for the issue now heavily depends on the changes patched by this PR, so I would be happy if you would merge this first if everything but character escaping is fine 🙇

src/cljam/io/vcf/writer.clj
```diff
@@ -49,64 +49,69 @@
     (cstr/upper-case (name k))
     (->camelCaseString k)))
 
+(defn- add-extra-fields
+  [fields m required-keys]
+  (reduce-kv (fn [fields k v]
+               (conj fields (str (->camelCaseString k) "=\"" v "\"")))
```
totakke: Isn't `->PascalCaseString` more suitable? I couldn't find a mention of case for keys of extra fields from the spec, but many VCF seem to use `PascalCaseKey=value` style, except for contig extra fields. I feel it is more conventional to use `PascalCaseKey` by default and `camelCaseKey` for contig fields.

athos: I agree. I'll update the code so that the keys for extra fields will be emitted in PascalCase. Thanks!

athos:
@totakke Thank you for reviewing this patiently 😭

> You can create a new branch from this branch if depending on this PR.

Sure, I will.
And now I have additional TODOs @alumi has pointed out for me offline, so please be patient for a little bit more until they are all done.

athos:
Added some more changes:

- An extra field key is now emitted in PascalCase
- Some relatively common optional fields are treated specially with prior knowledge on them (such as the fields `length` and `assembly` for `##CONFIG`, the fields `Genomes` and `Mixture` for `##SAMPLE`, etc.)
- The fields of `##PEDIGREE` meta info lines (except the `ID` field) are no longer treated as ordinary extra fields: They are treated as special optional fields now.
- The `idx` field is always uppercased on serialization

athos:
Added `##SAMPLE` and `##PEDIGREE` meta info lines into VCF/BCF files for testing.

I also tried to add `##META` lines, but bcftools doesn't seem to support them yet and couldn't generate a corresponding BCF file from VCF.

totakke:
Thank you for great enhancement and your effort!

athos:
Thank you for spending lots of time for reviewing and commenting this PR!
 I'm glad to see this feature merged 😆

PR #138, opened 2018-07-05
Character escaping support for string fields in VCF meta info

athos:
This PR provides character escaping support for VCF string fields (that was reported in #135).

With this change, you'll be able to use the characters `\` and `"` in VCF string fields by escaping them with `\`, such like `\\` and `\"`.

```clj
(require '[cljam.io.vcf.reader :as reader])
(#'reader/parse-structured-line "ID=ID,Description=\"\\\"This\\\" is a description\",Note=\"You can use \\\" in string fields by escaping it with \\\\\"")
;=> {:id "ID", :description "\"This\" is a description", :note "You can use \" in string fields by escaping it with \\"}
```

It also improves error checking for string field format:

```clj
(#'reader/parse-structured-line "x=\"unbalanced double quote")
;; RuntimeException Unexpected end of string field

(#'reader/parse-structured-line "x=\"unknown \\x escape sequence\"")
;; RuntimeException Either '\' or '"' was expected immediately after '\', but got 'x'
```

My only concern at the moment is that the VCF reader sometimes ends up making a weird result with no error if there are some redundant whitespaces around a comma (`,`):

```clj
(#'reader/parse-structured-line "x=\"foo\"  ,  y=\"bar\"")
;=> {:x "foo", :,-y "bar"}
```

I'm not sure if this form of meta info line is valid against the spec and wondering whether or not we should take a special care of those particular cases.

### DIFF
--- a/src/cljam/io/vcf/reader.clj
+++ b/src/cljam/io/vcf/reader.clj
@@ -64,10 +64,26 @@
         => {:id \"NS\" :number \"1\" :type \"Integer\"
             :description \"Number of Samples With Data\"}"
   [s]
-  (->> (re-seq #"([\w:/\.\?\-]*)=(?:(?:\"([^\"]*?)\")|([\w:/\.\?\-]*))" s)
-       (map (fn [[_ k v1 v2]]
-              {(->kebab-case-keyword k) (dot->nil (or v2 v1))}))
-       (apply merge)))
+  (letfn [(parse-string-field [s]
+            (loop [s s, acc []]
+              (if-let [[_ pre c post] (re-matches #"([^\\\"]*?)([\\\"])(.*)" s)]
+                (if (= c "\\")
+                  ;; expecting one of the special characters occurs immediately after \\
+                  (if (#{\\ \"} (first post))
+                    (recur (subs post 1) (conj acc pre (first post)))
+                    (let [msg (str "Either '\\' or '\"' was expected immediately after '\\', "
+                                   "but got '" (first post) "'")]
+                      (throw (RuntimeException. msg))))
+                  ;; c == \", that is, we are now at the end of the string field
+                  [nil (cstr/join (conj acc pre)) (cstr/replace post #"^," "")])
+                (throw (RuntimeException. "Unexpected end of string field")))))]
+    (loop [s s, m {}]
+      (if-let [[_ k more] (some->> s (re-matches #"([^=]+?)=(.*)"))]
+        (let [[_ v more] (if (= (first more) \")
+                           (parse-string-field (subs more 1))
+                           (re-matches #"([^,]*?)(?:,(.*))?" more))]
+          (recur more (assoc m (->kebab-case-keyword k) (dot->nil v))))
+        m))))
 
 (defn- parse-meta-info-contig
   [m]
--- a/src/cljam/io/vcf/writer.clj
+++ b/src/cljam/io/vcf/writer.clj
@@ -73,7 +73,7 @@
         (:assembly m) (conj (str "assembly=" (:assembly m)))
         (:md-5 m) (conj (str "md5=" (:md-5 m)))
         (:url m) (conj (str "URL=" (:url m)))
-        (:species m) (conj (str "species=\"" (:species m) "\""))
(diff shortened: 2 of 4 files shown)


alumi:
Thanks for your PR 👍 
Could you add these test codes (or some equivalent ones) to the project?

> ```clojure
> (#'reader/parse-structured-line "ID=ID,Description=\"\\\"This\\\" is a description\",Note=\"You can use \\\" in string fields by escaping it with \\\\\"")
> ;=> {:id "ID", :description "\"This\" is a description", :note "You can use \" in string fields by escaping it with \\"}
> 
> (#'reader/parse-structured-line "x=\"unbalanced double quote")
> ;; RuntimeException Unexpected end of string field
> 
> (#'reader/parse-structured-line "x=\"unknown \\x escape sequence\"")
> ;; RuntimeException Either '\' or '"' was expected immediately after '\', but got 'x'
> ```

athos:
Ah, I see. I will.

athos:
Just added the missing test cases ✌️

src/cljam/io/vcf/reader.clj
```diff
@@ -64,10 +64,26 @@
         => {:id \"NS\" :number \"1\" :type \"Integer\"
             :description \"Number of Samples With Data\"}"
   [s]
-  (->> (re-seq #"([\w:/\.\?\-]*)=(?:(?:\"([^\"]*?)\")|([\w:/\.\?\-]*))" s)
-       (map (fn [[_ k v1 v2]]
-              {(->kebab-case-keyword k) (dot->nil (or v2 v1))}))
-       (apply merge)))
+  (letfn [(parse-string-field [s]
```
alumi: Great implementation but it looks like a bit too complex to me 🤔 

Regexp like following will do the job except throwing the exceptions, right? or is it not enough?
```clojure
(->> (re-seq #"([\w:/\.\?\-]*)=(?:(?:\"((?:[^\\\"]|\\\"|\\\\)*)\")|([\w:/\.\?\-]*))\s*(?=(?:,|$))" s)
     (into {} (map (fn [[_ k v1 v2]]
                     [(->kebab-case-keyword k) (cstr/replace (dot->nil (or v2 v1)) #"\\(?=[\\\"])" "")]))))
```

totakke:
Doesn't `species` in contig line also need to be escaped...? It is not mentioned in the spec, but I think it should be escaped as long as it is wrapped by double-quote.

```clojure
(defn- stringify-meta-info-contig
  [m]
  (-> [(str "ID=" (:id m))]
      (cond->
        (:length m) (conj (str "length=" (:length m)))
        (:assembly m) (conj (str "assembly=" (:assembly m)))
        (:md-5 m) (conj (str "md5=" (:md-5 m)))
        (:url m) (conj (str "URL=" (:url m)))
        (:species m) (conj (str "species=\"" (:species m) "\"")) ;; <--
        (:taxonomy m) (conj (str "taxonomy=" (:taxonomy m)))
        (:idx m) (conj (str "IDX=" (:idx m))))
      (pack-meta-info m [:id :length :assembly :md-5 :url :species :taxonomy :idx])))
```

src/cljam/io/vcf/reader.clj
athos: > it looks like a bit too complex

Definitely. I think so, too 😂

As far as I'm aware, the only difference between your implementation and mine is just error checking, as you told. But, I'm suspecting the regex approach would scale enough to parse `##META` line in the future. I don't want to see that monolithic regex more complex :sob: 
I think the ideal implementation may be in between the two.

athos:
> It is not mentioned in the spec, but I think it should be escaped as long as it is wrapped by double-quote.

Ah, good catch! It should be, I think.

athos:
Added missing character escaping for `##CONTIG`'s `species`.

src/cljam/io/vcf/reader.clj
alumi: > But, I'm suspecting the regex approach would scale enough to parse ##META line in the future

OK, let's leave it as it is until the spec of `##META` line gets clarified.
Thanks!

totakke:
Thanks!

PR #146, opened 2018-08-31
Fix options of cljam.algo.pileup/pileup

alumi:
#### Summary
- Add an `ignore-overlaps?` option to `cljam.algo.pileup/pileup`
- Skip filtering bases when `min-base-quality` is ~zero~ not positive

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗

### DIFF
--- a/src/cljam/algo/pileup.clj
+++ b/src/cljam/algo/pileup.clj
@@ -195,12 +195,22 @@
       (step xs))))
 
 (defn mpileup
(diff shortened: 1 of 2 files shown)


alumi:
Added a minor change and amended the commit.

alumi:
@totakke Thanks for taking a look! I've added some docstrings.

PR #147, opened 2018-09-03
Fix CI config to execute tests in series

alumi:
Like [varity](https://github.com/chrovis/varity/blob/065c7e6fcdaabc129f4d218cf504bb44f3dc190d/.travis.yml#L15), reuse CI instances to reduce time to execute tests: 19.5min => 13min


alumi:
Thank you!! 😃

PR #148, opened 2018-09-04
Fix a bug in `cljam.algo.pileup/pileup`

alumi:
#### Summary
This PR fixes a bug in `cljam.algo.pileup/pileup`.

#### Problem

IllegalArgumentException in `cljam.algo.pileup/pileup`
```
java.lang.IllegalArgumentException: Value out of range for byte: 186
```

#### Cause
The `qual` field in `cljam.io.pileup.PileupBase` is declared as a `byte` because a phred-33 encoded base quality score never exceeds `93`:

https://github.com/chrovis/cljam/blob/02addc40def7515d2f30db8032fe4cf9776c5bd8/src/cljam/io/pileup.clj#L15

The record is also constructed in `cljam.algo.pileup`:

https://github.com/chrovis/cljam/blob/02addc40def7515d2f30db8032fe4cf9776c5bd8/src/cljam/algo/pileup.clj#L85-L94

But quality scores can be `128` ~ `200` while piling alignments up with the option `:ignore-overlaps false`:

https://github.com/chrovis/cljam/blob/02addc40def7515d2f30db8032fe4cf9776c5bd8/src/cljam/algo/pileup.clj#L117-L120

#### Changes
- Use `short` primitive type hinting instead of `byte` in `cljam.io.pileup.PileupBase`
- Fix broken tests related to `pileup`

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗

### DIFF
--- a/src/cljam/io/pileup.clj
+++ b/src/cljam/io/pileup.clj
@@ -12,7 +12,7 @@
 (defrecord PileupBase [^boolean start?
                        mapq ;; byte?
                        ^char base
-                       ^byte qual
+                       ^short qual
                        ^boolean reverse?
                        ^boolean end?
                        insertion ;; String?
--- a/test/cljam/algo/pileup_test.clj
+++ b/test/cljam/algo/pileup_test.clj
@@ -252,9 +252,9 @@
 (def ^:private reads-for-pileup
   (mapv
    p/map->SAMAlignment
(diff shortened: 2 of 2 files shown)


alumi:
@athos @totakke @r6eve Thanks for your review! 😸

PR #151, opened 2018-10-29
Fix some fields of cljam.io.pileup LocusPile and PileupBase

alumi:
#### Summary
For clarification and better performance,
- Remove `count` and `ref` fields from `defrecord` of `cljam.io.pileup.LocusPile`
    - optional for `cljam.io.pileup.PileupReader`
- Always fill `mapq` of `cljam.io.pileup.PileupBase` when available

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ cache:
   directories:
     - $HOME/.m2
 jdk:
-  - oraclejdk10
   - oraclejdk9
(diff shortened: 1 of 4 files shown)


alumi:
@totakke @athos @r6eve Thank you for reviewing!! 😸

PR #154, opened 2018-11-12
Remove normalization of chromosome names from BED reader

alumi:
#### Summary

**BREAKING CHANGE**
Decouple chromosome name normalization from BED i/o.
Chromosome name normalization is an irreversible transformation and should not be complected.

#### Changes

- removed `cljam.util.chromosome/normalize-chromosome-key` used in `cljam.io.bed/read-fields`
- removed `cljam.io.bed/read-raw-fields` and `cljam.io.bed/write-raw-fields`which are no longer necessary

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗
- `lein eastwood` 🆗

### DIFF
--- a/src/cljam/io/bed.clj
+++ b/src/cljam/io/bed.clj
@@ -154,7 +154,6 @@
   This function converts the coordinate into cljam style: 1-origin and inclusive-start / inclusive-end."
   [m]
   (-> m
-      (update :chr chr/normalize-chromosome-key)
       (update :start inc)
       (update-some :thick-start inc)))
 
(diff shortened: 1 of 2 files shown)


totakke:
There's no definite deprecation policy for now, but I'm careful about versioning in a case where breaking changes are included. #144 already included breaking change, so I think breaking change in this PR is not so serious matter. The next version will be 0.7.0 (not 0.6.1) in any case.

I personally feel we don't have to care about graceful deprecation and serious backward compatibility because cljam is not totally stable version (>= 1.0.0) yet. I want to hear you all opinions on that point.

totakke:
@athos Thanks!

I also think deprecation warning should be marked if possible. However, this shall not apply when deprecation marking is not easy. I feel that stance is enough for now.

As I commented above, the warnings are not needed about this PR.

alumi:
Thanks! 😸

PR #156, opened 2018-11-21
Remove useless nil entries in variants

### LINKED ISSUE #155: Seemingly useless nil entry occasionally included in VCF parsing results
When parsing a VCF file just having the 8 required fields (ie. `chrom`/`pos`/`id`/`ref`/`alt`/`qual`/`filter`/`info`), the resulting maps contain a seemingly useless `nil nil` entry as follows:

```clj
(def r (vcf/reader "/path/to/file.vcf"))
(def vs (vcf/read-variants r))
(first vs)
;=> {nil nil, :alt ["C"], :ref "T", :pos 123456, :filter (:PASS), :id nil, :info {:DP 40, :RD 31, :AD (9), :AF (0.225)}, :qual nil, :chr "chr1"}
;    ^^^^^^^
```

If the VCF file has some extra fields other than the required ones, it dosen't happen.

alumi:
#### Summary
This PR fixes #155 (Thanks for the ticket, @athos!)

#### Problem
- See #155 

#### Cause
- `update` / `assoc` on `nil` key-values

#### Changes
- Check the header columns before `update` / `assoc`
- I also fixed another bug that BCFReader fails parsing `:qual` value

#### Affects
- VCF reader
- BCF reader

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗
- `lein eastwood` 🆗


alumi:
@totakke @athos @r6eve  Thank you for reviewing! 😃

PR #162, opened 2019-05-17
Add utility functions for manipulating genotypes

alumi:
#### Summary
This PR adds 4 utility functions related to genotype ordering in VCF.
I've refined the implementation a bit.
Basically, these are useful for working with genotype information values defined as `Number=G` like genotype likelihoods `GL`.
Such values should be provided for all possible genotypes with given ploidy and number of alleles (`REF` + `ALT`).
For more details, please refer to the [spec "VCFv4.3 1.6.2 Genotype fields"](https://github.com/samtools/hts-specs/blob/d962e2fc9e2456c737a91e88cbc40ccf4a732650/VCFv4.3.tex#L442).

- `genotype-seq`: list all genotypes
- `genotype-index`: get an index for a genotype
- `biallelic-genotype`: multiallelic genotype string to biallelic one
- `biallelic-coll`: multiallelic genotype-wise values to biallelic one

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗
- `lein cloverage` 🆗

PR #163, opened 2019-05-17
Add support for breakends in VCF

alumi:
#### Summary

This PR adds basic support for breakends in VCF.
Breakends are variant records which have `SVTYPE=BND` in the `INFO` field.
They have a dedicated format of string representation for `ALT` field.
For more details, please refer to the [spec "5.4 Specifying complex rearrangements with breakends"](https://github.com/samtools/hts-specs/blob/d962e2fc9e2456c737a91e88cbc40ccf4a732650/VCFv4.3.tex#L859)

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗
- `lein cloverage` 🆗

### DIFF
--- a/src/cljam/io/vcf/util.clj
+++ b/src/cljam/io/vcf/util.clj
@@ -1,5 +1,6 @@
 (ns cljam.io.vcf.util
-  (:require [clojure.string :as cstr]))
+  (:require [clojure.string :as cstr]
+            [proton.core :as p]))
 
 (definline dot-or-nil?
   "Checks if given string is equal to \".\" or nil."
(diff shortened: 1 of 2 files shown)


alumi:
Resolved conflicts with #162

src/cljam/io/vcf/util.clj
```diff
@@ -233,3 +234,54 @@
           (update :info stringify-info)
           (update fmt-kw stringify-format)
           (merge (into {} (for [k sample-kws] [k (stringify-sample (v fmt-kw) (v k))])))))))
+
+(def ^:private long-breakend-regexp
+  ;;     pre-seq      [ or ]  chr      pos   [ or ]    post-seq
+  #"((?:[ACGTN]*|\.))([\[\]])(.+?)(?::(\d+))([\[\]])((?:[ACGTN]*|\.))")
```
r6eve: This regexp is capturing `pre-seq` and `post-seq` after non-capturing, it can be written the following directly.

```clojure
;; pre-seq      [ or ]  chr      pos    [ or ]  post-seq
#"([ACGTN]*|\.)([\[\]])(.+?)(?::(\d+))?([\[\]])([ACGTN]*|\.)"
```

alumi: Fixed in 51ac859. Thanks!

src/cljam/io/vcf/util.clj
```diff
@@ -233,3 +234,54 @@
...
+
+(defn parse-breakend
+  "Parses an ALT allele string of SVTYPE=BND.
+  Returns a map with mandatory keys:
+  - `:bases` bases that replaces the reference place
+  - `:join` `:before` or `:after`
+  and optional keys for a mate:
+  - `:chr` chromosome name of the mate sequence
+  - `:pos` genomic position of the mate sequence
+  - `:strand` strand of the mate sequence
```
athos: I'm not too confident, but I'm wondering if `strand` is a little confusing name here because it does not necessarily match the physical strand of the mate sequence (is that correct?).
It appears to me the VCF spec carefully uses the word `orientation` instead of `strand` for this meaning. What do you think of it?

alumi: As you mentioned, `orientation` is clearly distinguished from `:strand` in the VCF spec.

> ```
> REF	ALT		Meaning
> s		t[p[	piece extending to the right of p is joined after t
> s		t]p]	reverse comp piece extending left of p is joined after t
> s		]p]t	piece extending to the left of p is joined before t
> s		[p[t	reverse comp piece extending right of p is joined before t
> ```

Summarizing the above:

|ALT|`:strand`|orientation|`:join`|
|:-:|:-:|:-:|:-:|
|`t[p[`|`:forward`|right|`:after`|
|`t]p]`|`:reverse`|left|`:after`|
|`]p]t`|`:forward`|left|`:before`|
|`[p[t`|`:reverse`|right|`:before`|

I chose to omit the `:orientation` since one of the three flags is redundant for four possible modes.

src/cljam/io/vcf/util.clj
```diff
@@ -233,3 +234,54 @@
...
+                     (if post :forward :reverse)
+                     (if post :reverse :forward))})))
+    (when-let [[_ [pre-dot] bases [post-dot]]
+               (re-matches short-breakend-regexp alt)]
+      (when (and (or pre-dot post-dot)
+                 (not (and pre-dot post-dot)))
+        {:bases bases, :join (if pre-dot :before :after)}))))
+
+(defn stringify-breakend
+  "Returns a string representation of a breakend."
```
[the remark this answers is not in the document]
alumi: Fixed in d848699. Thanks!

src/cljam/io/vcf/util.clj
athos: Thank you for elaborating on the difference between strand and orientation! That does make much sense.

alumi:
Thanks!! 🎉

PR #164, opened 2019-05-20
Add a function to inspect an ALT allele of VCF

alumi:
~~**This PR is depending on #163**~~

#### Summary
This PR adds a function to inspect an `ALT` allele of a variant in VCF.
The `ALT` field of VCF is so expressive that it is not so straightforward to know what they mean.
`cljam.io.vcf.util/inspect-allele` provides a way to classify `ALT` alleles into ~~9~~ ~~10~~ 13 types.
With this feature, we can easily transform VCF variants based on their types. (e.g. filtering only SNPs)

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗
- `lein cloverage` 🆗

### DIFF
--- a/src/cljam/io/vcf/util.clj
+++ b/src/cljam/io/vcf/util.clj
@@ -312,7 +312,7 @@
 
       (and (< left-match ref-length) (= left-match alt-length))
       {:type :deletion, :offset (dec left-match),
-       :n-bases (- left-match ref-length), :deleted (subs ref left-match)}
+       :n-bases (- ref-length left-match), :deleted (subs ref left-match)}
 
       (= (inc matched-length) ref-length alt-length)
       {:type :snv, :ref (nth ref left-match),
@@ -320,7 +320,7 @@
 
       (= matched-length alt-length)
       {:type :deletion, :offset (dec left-match),
-       :n-bases (- left-match (- ref-length right-match)),
+       :n-bases (- ref-length right-match left-match),
        :deleted (subs ref left-match (- ref-length right-match))}
 
       (= matched-length ref-length)
--- a/src/cljam/io/vcf/util.clj
+++ b/src/cljam/io/vcf/util.clj
@@ -304,8 +304,10 @@
                          (Math/min (- (Math/min ref-length alt-length)
                                       left-match)))
         matched-length (+ left-match right-match)]
-    ;; Note that `{:type :ref}` is handled in the caller
     (cond
+      (= left-match ref-length alt-length)
+      {:type :ref}
+
       (and (= left-match ref-length) (< left-match alt-length))
       {:type :insertion, :offset (dec left-match),
        :n-bases (- alt-length left-match), :inserted (subs alt left-match)}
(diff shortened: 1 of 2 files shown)


alumi:
Rebased the branch since #163 is merged

alumi:
Sorry. Added a commit 2d1b506 to fix a bug.

src/cljam/io/vcf/util.clj
```diff
@@ -286,3 +286,93 @@
...
+        matched-length (+ left-match right-match)]
+    ;; Note that `{:type :ref}` is handled in the caller
+    (cond
+      (and (= left-match ref-length) (< left-match alt-length))
+      {:type :insertion, :offset (dec left-match),
+       :n-bases (- alt-length left-match), :inserted (subs alt left-match)}
+
+      (and (< left-match ref-length) (= left-match alt-length))
+      {:type :deletion, :offset (dec left-match),
+       :n-bases (- left-match ref-length), :deleted (subs ref left-match)}
```
r6eve: I think `:n-bases` means the number of inserted/deleted bases, but it's bounded minus value (i.e. non-natural number) when `:type` is `:deletion`. Maybe it's more appropriate to make `:n-bases` take only natural numbers.

src/cljam/io/vcf/util.clj
```diff
@@ -286,3 +286,93 @@
...
+      (re-matches #"(?i)([ACGTN]<.+>|<.+>[ACGTN])" alt)
+      (if (cstr/starts-with? alt "<")
+        {:type :complete-insertion,
+         :join :before,
+         :base (last alt),
+         :id (subs alt 1 (- alt-length 2))}
+        {:type :complete-insertion,
+         :join :after,
+         :base (first alt),
+         :id (subs alt 2 (dec alt-length))})
```
r6eve: I think it's a better way to use capturing values in regexp (please correct me if I'm wrong).

```clojure
  [ref alt]
  (if (or (#{"." "*" "X" "<*>" "<X>"} alt)
          (= (cstr/upper-case ref) (cstr/upper-case alt)))
    {:type :ref}
    (condp re-matches alt
      #"<(.+)>" :>> (fn [[_ id]] {:type :id, :id id})
      #"(?i)<(.+)>([ACGTN])" :>> (fn [[_ id [base]]]
                                   {:type :complete-insertion,
                                    :join :before,
                                    :base base,
                                    :id id})
      #"(?i)([ACGTN])<(.+)>" :>> (fn [[_ [base] id]]
                                   {:type :complete-insertion,
                                    :join :after,
                                    :base base,
                                    :id id})
      (or (some-> (parse-breakend alt)
                  (assoc :type :breakend))
          (if (and (re-matches #"(?i)[ACGTN]+" ref)
                   (re-matches #"(?i)[ACGTN]+" alt))
            (inspect-nucleotides-allele ref alt)
            {:type :other})))))
```

src/cljam/io/vcf/util.clj
alumi: Thanks! Fixed in 07aa4d7

src/cljam/io/vcf/util.clj
alumi: Thank you for suggesting `condp`!
 I rewrote the patch with it and also tweaked the categorization in dc63967.
I think it looks better now 😄

alumi:
Thank you for reviewing! @r6eve @athos

PR #166, opened 2019-05-23
Normalize variants in a VCF file

alumi:
~~**This PR is depending on #164**~~

#### Summary
Add `cljam.io.vcf.util.normalize/normalize` which converts variants into canonical forms.
cf. `bcftools norm -m-`

See the test cases for typical conversion examples.

```
ref0: TATTTTTTTTTTTT TC  TTTTT
alt1:                TTCC     <- not right-trimmed
alt2:                TTC      <- not right-trimmed

ref0: TATTTTTTTTTTTT T  CTTTTT
alt1:                TTC
alt2:                TT       <- not left-aligned

ref0: T A TTTTTTTTTTTT T  CTTTTT
alt1:                  TTC
alt2:   AT
```

**input** multiallelic, not left-aligned, not right-trimmed
https://github.com/chrovis/cljam/blob/41881832c20f96c25d91027a73cab09a34378950/test/cljam/io/vcf/util/normalize_test.clj#L328-L330

**output** biallelic, left-aligned, right-trimmed
https://github.com/chrovis/cljam/blob/41881832c20f96c25d91027a73cab09a34378950/test/cljam/io/vcf/util/normalize_test.clj#L313-L318

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗
- `lein cloverage` 🆗
- `lein eastwood` 🆗

### DIFF
--- a/src/cljam/io/vcf/util/normalize.clj
+++ b/src/cljam/io/vcf/util/normalize.clj
@@ -111,9 +111,7 @@
         (-> (if (integer? END)
               (assoc-in v [:info :END] (dec (+ pos' (count ref'))))
               v)
-            (assoc :pos pos')
-            (assoc :ref ref')
-            (assoc :alt alt')))
+            (assoc :pos pos' :ref ref' :alt alt')))
       v)))
 
 (defn- trim-left
--- a/src/cljam/io/vcf/util/normalize.clj
+++ b/src/cljam/io/vcf/util/normalize.clj
@@ -46,14 +46,6 @@
                           [s (biallelic-map (v s) format-id->number i n)]))
                        samples))))))))))
 
-(defn same-ref?
-  "Checks if a REF allele matches the reference sequence."
-  [seq-reader {:keys [chr ^long pos ^String ref]}]
-  (let [ref-len (.length ref)
-        ref-region {:chr chr :start pos :end (dec (+ pos ref-len))}
-        ref-seq (io-seq/read-sequence seq-reader ref-region {:mask? false})]
-    (and ref-seq (.equalsIgnoreCase ^String ref-seq ref))))
-
 (defn- regions-before [chr ^long pos ^long window]
   (->> (range pos 1 (- window))
        (map
--- a/src/cljam/io/vcf/util/normalize.clj
+++ b/src/cljam/io/vcf/util/normalize.clj
@@ -62,15 +62,15 @@
            :start (Math/max 1 (- e (dec window))),
            :end (dec e)}))))
 
-(defn- char-equals-ignore-case
+(defn- char-equals-ignore-case?
   ([_]
    true)
   ([x y]
    (= (Character/toUpperCase ^char x)
       (Character/toUpperCase ^char y)))
   ([x y & z]
-   (and (char-equals-ignore-case x y)
-        (apply char-equals-ignore-case y z))))
+   (and (char-equals-ignore-case? x y)
+        (apply char-equals-ignore-case? y z))))
 
 (defn- trim-right
   "Trims all duplicated allele sequences from right. If reached the start,
@@ -90,7 +90,7 @@
                   regs)
         allele-seqs (map #(concat (cstr/reverse %) ref-seqs) alleles)
         matched-length (->> allele-seqs
-                            (apply map char-equals-ignore-case)
+                            (apply map char-equals-ignore-case?)
                             (take-while true?)
                             count
                             long
@@ -121,8 +121,7 @@
   [{:keys [ref alt] :as v}]
   (let [alleles (cons ref alt)
         n-trim-left (->> alleles
-                         (map cstr/upper-case)
-                         (apply map =)
+                         (apply map char-equals-ignore-case?)
                          (take-while true?)
                          count)
         min-len (long (apply min (map count alleles)))
--- a/src/cljam/io/vcf/util/normalize.clj
+++ b/src/cljam/io/vcf/util/normalize.clj
@@ -108,10 +108,8 @@
                                  cstr/join)
                            alleles
                            allele-seqs)]
-        (-> (if (integer? END)
-              (assoc-in v [:info :END] (dec (+ pos' (count ref'))))
-              v)
-            (assoc :pos pos' :ref ref' :alt alt')))
+        (cond-> (assoc v :pos pos' :ref ref' :alt alt')
+          (integer? END) (assoc-in [:info :END] (dec (+ pos' (count ref'))))))
       v)))
 
 (defn- trim-left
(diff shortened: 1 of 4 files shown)


alumi:
Rebased

src/cljam/io/vcf/util/normalize.clj
```diff
@@ -0,0 +1,176 @@
...
+
+(defn- regions-before [chr ^long pos ^long window]
+  (->> (range pos 1 (- (dec window)))
+       (map
+        (fn [^long e]
+          {:chr chr,
+           :start (Math/max 1 (- e (dec window))),
+           :end (dec e)}))))
+
+(defn- char-equals-ignore-case
```
r6eve: This function is a predicate. Can you make it end in a question mark?
https://github.com/bbatsov/clojure-style-guide#pred-with-question-mark

src/cljam/io/vcf/util/normalize.clj
```diff
@@ -0,0 +1,176 @@
...
+                                 reverse
+                                 cstr/join)
+                           alleles
+                           allele-seqs)]
+        (-> (if (integer? END)
+              (assoc-in v [:info :END] (dec (+ pos' (count ref'))))
+              v)
+            (assoc :pos pos')
+            (assoc :ref ref')
+            (assoc :alt alt')))
```
r6eve: `(assoc map key val)` is also `(assoc map key val & kvs)`, so it can be written as:

```clojure
(assoc :pos pos' :ref ref' :alt alt')
```

src/cljam/io/vcf/util/normalize.clj
```diff
@@ -0,0 +1,176 @@
...
+      v)))
+
+(defn- trim-left
+  "Trims all duplicated allele sequences from left."
+  [{:keys [ref alt] :as v}]
+  (let [alleles (cons ref alt)
+        n-trim-left (->> alleles
+                         (map cstr/upper-case)
+                         (apply map =)
```
r6eve: This two lines are equivalent to `(apply map char-equals-ignore-case)`.

alumi:
@r6eve Thanks for the comments! Fixed in 6b3eece and b634132.

src/cljam/io/vcf/util/normalize.clj
```diff
@@ -0,0 +1,173 @@
...
+  ([seq-reader variant]
+   (realign seq-reader variant {}))
+  ([seq-reader
+    {:keys [ref alt] :as variant}
+    {:keys [window] :or {window 100}}]
+   (cond
+     (and
+      ref
+      (seq alt)
+      (re-matches nucleotides-regexp ref)
```
r6eve: I think the check against `ref` is a little bit redundant because `vcf-util/inspect-allele` can also play the role.
https://github.com/chrovis/cljam/blob/55353b7bee71cda5313655d2c91970094a35c728/src/cljam/io/vcf/util.clj#L358-L373

alumi: That's a good point. Thanks! Fixed in 9e05452

src/cljam/io/vcf/util/normalize.clj
```diff
@@ -0,0 +1,165 @@
...
+                     pos)
+            [ref' & alt'] (map
+                           #(->> %2
+                                 (take (inc (- (dec (+ pos (count %1))) pos')))
+                                 (drop matched-length)
+                                 reverse
+                                 cstr/join)
+                           alleles
+                           allele-seqs)]
+        (-> (if (integer? END)
```
athos: Like [L149](https://github.com/chrovis/cljam/pull/166/files#diff-4254e4fff1284962ff6e63264acfbd1fR149), these lines can be written as:

```clojure
(cond-> (assoc v :pos pos' :ref ref' :alt alt')
  (integer? END)
  (assoc-in [:info :END] (dec (+ pos' (count ref')))))
```

src/cljam/io/vcf/util/normalize.clj
```diff
@@ -0,0 +1,165 @@
...
+
+(defn same-ref?
+  "Checks if a REF allele matches the reference sequence."
+  [seq-reader {:keys [chr ^long pos ^String ref]}]
+  (let [ref-len (.length ref)
+        ref-region {:chr chr :start pos :end (dec (+ pos ref-len))}
+        ref-seq (io-seq/read-sequence seq-reader ref-region {:mask? false})]
+    (and ref-seq (.equalsIgnoreCase ^String ref-seq ref))))
+
+(defn- regions-before [chr ^long pos ^long window]
```
athos: I'm not sure if I read what you meant correctly, but it looks `regions-before` generates `window - 1` bases wide window of regions.

```clojure
(regions-before "chr1" 10 3)
;=> ({:chr "chr1", :start 8, :end 9}
;    {:chr "chr1", :start 6, :end 7}
;    {:chr "chr1", :start 4, :end 5}
;    {:chr "chr1", :start 2, :end 3}
;    {:chr "chr1", :start 1, :end 1})
```

src/cljam/io/vcf/util/normalize.clj
alumi: Thanks! Fixed in 595435f.

src/cljam/io/vcf/util/normalize.clj
alumi: Thank you for pointing that out! You're right, they should span exactly `window` bases. Fixed the function and test cases in f21905b.

src/cljam/io/vcf/util/normalize.clj
```diff
@@ -0,0 +1,163 @@
...
+                  (-> v
+                      (assoc :alt [allele])
+                      (update :info biallelic-map info-id->number i n)
+                      (into
+                       (map
+                        (fn [s]
+                          [s (biallelic-map (v s) format-id->number i n)]))
+                       samples))))))))))
+
+(defn same-ref?
```
athos: This function doesn't seem to be used anywhere in this namespace. Is it necessary to be defined here?

alumi: No, not necessary. Do you mean it should be in `cljam.io.vcf.util`?

athos: Yeah, I think it would suit better to `cljam.io.vcf.util` if it is not specific to VCF normalization.

alumi: Considering that current `cljam.io.vcf.util` is a namespace for VCF/BCF and that `same-ref?` compares two files of different types, how about making a new namespace like `cljam.io.vcf.util.check`?

athos: That sounds good, too 👍

alumi:
Thank you for your helpful feedback @r6eve @athos! 🙏

PR #167, opened 2019-06-20
Fix I/O of variants with empty info/individual fields

alumi:
#### Summary
This PR fixes three minor bugs in VCF/BCF I/O

1. Writing `:info {}` does not emit `.` (VCF)
   https://github.com/chrovis/cljam/blob/f91f8164db66ec89ac67cf2b298dfff9a7c42179/test/cljam/io/vcf/writer_test.clj#L50-L51
2. Writing a variant map without samples will cause a NPE (VCF)
   https://github.com/chrovis/cljam/blob/f91f8164db66ec89ac67cf2b298dfff9a7c42179/test/cljam/io/vcf/writer_test.clj#L62-L63
3. Omitted genotype fields are totally ignored (VCF/BCF)
   e.g. Reading `DP:HQ	4:51,51	2` does not assoc `:HQ nil` to the 2nd sample.
   https://github.com/chrovis/cljam/blob/f91f8164db66ec89ac67cf2b298dfff9a7c42179/test/cljam/test_common.clj#L711-L714

#### Tests

- `lein check` 🆗
- `lein test :all` 🆗
- `lein cloverage` 🆗
- `lein eastwood` 🆗

### DIFF
--- a/src/cljam/io/vcf/writer.clj
+++ b/src/cljam/io/vcf/writer.clj
@@ -32,7 +32,7 @@
 
 (defn- nil->dot
   [s]
-  (if (nil? s) "." s))
+  (or s "."))
 
 (defn- write-line
   [^BufferedWriter bwtr ^String s]
--- a/src/cljam/io/vcf/writer.clj
+++ b/src/cljam/io/vcf/writer.clj
@@ -32,7 +32,7 @@
 
 (defn- nil->dot
   [s]
-  (or s "."))
+  (if (nil? s) "." s))
 
 (defn- write-line
   [^BufferedWriter bwtr ^String s]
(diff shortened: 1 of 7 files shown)


src/cljam/io/vcf/writer.clj
```diff
@@ -32,7 +32,7 @@
 
 (defn- nil->dot
   [s]
-  (if (nil? s) "." s))
+  (or s "."))
```
r6eve: By the rewrite, it will also convert `false` to dot (`"."`). I think it's a good thing to undo the code or change the function name as something like `default-dot-value`. It's only a nitpick.

alumi: Though there's no concept of boolean values in the current version of VCF/BCF spec, I agree with you that it's misleading. Reverted in a455eac.

alumi:
Thank you for reviewing @athos @r6eve! :smiley:

PR #171, opened 2019-07-11
Add content-based file type detection

alumi:
#### summary

`cljam.io.sam/reader`, `cljam.io.vcf/reader` and `cljam.io.sequence/reader` determines a file type based on a suffix of the input file. This can be problematic when the suffix is not available. (e.g. hashed filenames for temporary use)
This PR adds fallbacks for detecting file types.
When failed to detect a file type from the suffix, those functions now try to detect it based on the content of the file.

#### tests
- `lein check` 🆗 
- `lein test :all` 🆗 
- `lein cloverage` 🆗 
- `lein eastwood` 🆗


test/cljam/io/util_test.clj
```diff
@@ -57,24 +59,80 @@
...
     "foo.BW" :bigwig)
   (are [?dir]
-      (are [?path] (thrown? Exception (io-util/file-type (str ?dir ?path)))
-        "foo.bam.gz"
-        "foo.SAM.gz"
-        "foo.cram"
-        "foo.CRAM"
-        "foo.2bit.gz"
-        "foo.bcf.gz")
+       (are [?path] (thrown? Exception (io-util/file-type (str ?dir ?path)))
```
r6eve: I think there's a path bug. When `?dir` is bound to `"/home/bar"` and `?path` is `"foo.bam.gz"`, `(str ?dir ?path)` becomes `/home/barfoo.bam.gz`. It's not an intentional result. Replacing `str` with `cio/file` might be the fix.

src/cljam/io/util.clj
```diff
@@ -156,9 +157,53 @@
...
+        #"^##gff-version 3" :gff
+        #"^@HD\t" :sam
+        #"^@SQ\t" :sam
+        #"^@RG\t" :sam
+        #"^@PG\t" :sam
+        #"^@CO\t" :sam
+        #"^>\S" :fasta
+        #"^@\S" :fastq
+        #"(variable|fixed)Step" :wig
+        #"^\S+\t\d+\t\d+\t\d+\t\d+\n" :fai
```
r6eve: Nitpick: I think the pattern could be concise a little: `#"^\S+(?:\t\d+){4}\n" :fai`

test/cljam/io/util_test.clj
alumi: Good catch! Fixed in 112988f

src/cljam/io/util.clj
alumi: Fixed in d2165e4

src/cljam/io/util.clj
```diff
@@ -156,9 +157,53 @@
...
+        #"^##fileformat=VCF" :vcf
+        #"^##gff-version 3" :gff
+        #"^@HD\t" :sam
+        #"^@SQ\t" :sam
+        #"^@RG\t" :sam
+        #"^@PG\t" :sam
+        #"^@CO\t" :sam
+        #"^>\S" :fasta
+        #"^@\S" :fastq
+        #"(variable|fixed)Step" :wig
```
athos: Just to confirm, is it intentional the leading `^` is omitted here?

src/cljam/io/util.clj
```diff
@@ -156,9 +157,53 @@
...
+  "Tries to detect a file format based on contents of the byte array `ba`.
+  The input byte array must be larger than 4 bytes. Note that detection of some
+  formats `#{:fasta :fastq :wig :fai :bed}` is based on naive heuristics and
+  thus can fail."
+  [^bytes ba]
+  (let [s (String. ba 0 (Math/min (int 64) (alength ba)))
+        i (-> (ByteBuffer/wrap ba)
+              (.order ByteOrder/LITTLE_ENDIAN)
+              (.getInt)
+              (bit-and 0xffffffff))]
```
athos: I didn't know this trick to coerce signed int to unsigned long 😲
But `Integer/toUnsignedLong` is a little more readable for this very purpose...?

src/cljam/io/util.clj
alumi: Yes, that's intentional. 

Wiggle files can be started with one of the followings:
- (optional) comment `#...`
- (optional) browser directive `browser ...`
- (optional) track definition `track ...`
- step declaration `variableStep ...`, `fixedStep ...`

First three are not specific to the format. Comments `#...` can be seen in VCF, `browser` and `track` are in BED.

[It seems like](http://genome.ucsc.edu/goldenPath/help/wiggle) wiggle format allows leading whitespaces for data lines. I'm not sure that leading whitespaces of step declaration lines are also permitted but our current implementation [supports](https://github.com/chrovis/cljam/blob/3c0371e9892f5541ebc91a61168bf29e7ecfba94/src/cljam/io/wig.clj#L83) them.

src/cljam/io/util.clj
alumi: Thanks! Fixed in 2c2123f.

src/cljam/io/util.clj
athos: > I'm not sure that leading whitespaces of step declaration lines are also permitted but our current implementation supports them.

Makes sense. OK, then 👍

r6eve:
LGTM. Thank you for the feature.

alumi:
Thank you for reviewing @athos @r6eve! 😺

PR #172, opened 2019-07-16
Fix nonintentional paths

r6eve:
`temp-dir` has a type of `java.lang.String` and the end is not `"/"`, so `"/"` is needed when concatenating `temp-dir` with files.
This PR fixes nonintentional paths that do not contain `"/"`, does not change the test results.


r6eve:
Thanks for the review.

PR #175, opened 2019-08-21
Fix reflection warnings

r6eve:
This PR avoids all reflections except those of external libraries when running `lein with-profile +dev test`.


alumi:
Thanks :v:

r6eve:
Thanks for the reviews!

PR #178, opened 2019-09-10
Fix creating temp directory to support multiple users

r6eve:
In the current implementation of creating temporary user's directory, there are two problems: always creates the directory of the same name `cljam`, and doesn't delete it after all is done. That might cause `Permission denied` if multiple users run cljam on the same machine.

This PR fixes the behavior; `with-temp-dir` macro creates an unique directory, and finally deletes it. Also, fixes a path to temporary directory such that the test code creates user's directory instead of test's directory.

### DIFF
--- a/src/cljam/algo/level.clj
+++ b/src/cljam/algo/level.clj
@@ -40,10 +40,10 @@
 
 (defn- cache-path
   "Create a path to the cache file."
-  [rdr i]
+  [rdr temp-dir i]
   (->> i
        (format "%s_%06d.bam" (util/basename (protocols/reader-url rdr)))
-       (cio/file util/temp-dir)
+       (cio/file temp-dir)
        (.getCanonicalPath)))
 
 (defn- add-level-mt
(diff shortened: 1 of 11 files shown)


src/cljam/util.clj
```diff
@@ -2,17 +2,35 @@
...
+    (zero? (count bindings)) `(do ~@body)
+    (symbol? (bindings 0)) (let [[dir prefix] bindings]
+                             `(let [~dir (->> 0
+                                              (make-array FileAttribute)
+                                              (Files/createTempDirectory ~prefix)
+                                              .toFile)]
+                                (try
+                                  (with-temp-dir ~(subvec bindings 2) ~@body)
+                                  (finally
+                                    (doseq [f# (.listFiles ~dir)]
```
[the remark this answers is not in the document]
alumi: Also, please add some test cases for this macro to check that:

- Temporary directories are created, named with given prefixes
- These directories are deleted outside of the macro
- Raises no exceptions when the directories are deleted before entering the `finally` clause
- Deletes all subdirectories created inside the macro by its user
- Deletes just symbolic links and keep their originals if they're out of the temporary directories

src/cljam/util.clj
```diff
@@ -2,17 +2,45 @@
...
 
 ;; Disk cache
 ;; ----------
 
-(def temp-dir (let [system-tmp-dir-path (or (System/getenv "TMPDIR")
-                                            (System/getProperty "java.io.tmpdir"))
-                    dir-path (.getPath (file system-tmp-dir-path "cljam"))]
-                (.mkdirs (file dir-path))
-                dir-path))
+(defmacro with-temp-dir
```
athos: The expanded form of this macro looks somewhat large. More importantly, it contains a `reify` form in it, which would redundantly generate a dedicated anonymous class for each macro callsite.

How about breaking the implementation down into some functions like `create-temp-dir` and `delete-temp-dir`, and using them in the expanded form?

r6eve:
Thank you for the helpful suggestion! I didn't know that.
To avoid the costs, I decided to make two functions and use them in expanded forms. 48205643031551d63e6f0f68a1c64df3f216fe69

athos:
Ah, sorry if I confused you 🙇 

 I think providing `with-temp-dir` as an internal utility is completely fine. By "using them (the functions) in the expanded form", I just meant something like:

```clojure
(defmacro with-temp-dir [bindings & body]
  (cond ...
        (symbol? (bindings 0))
        (let [[dir prefix] bindings]
          `(let [~dir (create-temp-dir ~prefix)]
             (try
               (with-temp-dir ~(subvec bindings 2) ~@body)
               (finally
                 (delete-temp-dir! ~dir)))))
        ...))
  ```

File: src/cljam/algo/level.clj [hunk 3]
```diff
@@ -54,23 +54,24 @@
...
-                                   (sam/read-alignments r)
-                                   (map (partial add-level! (volatile! [])))
-                                   (#(sam/write-alignments w % hdr))))
-                            cache))]
-          ;; merge
-          (doseq [cache xs]
-            (with-open [r (sam/reader cache)]
-              (sam/write-blocks wtr (sam/read-blocks r)))))))))
+      ;; split and compute levels
+      (let [temp-dir (util/create-temp-dir "cljam.also.level")
```
alumi: `"cljam.algo.level"`

src/cljam/algo/sorter.clj
```diff
@@ -150,17 +143,19 @@
...
-                       protocols/reader-url
-                       util/basename
-                       (partial gen-cache-filename (name cache-fmt) temp-dir))
-          key-fn ({header/order-coordinate (partial coordinate-key (refmap (sam/read-refs rdr)))
-                   header/order-queryname queryname-key} mode)
-          splitted-files (split* rdr chunk-size name-fn mode #(sort-by-index key-fn %))
-          hdr (header/sorted-by mode (header/update-version (sam/read-header rdr)))]
-      (logging/info (str "Merging from " (count splitted-files) " files..."))
-      (merge* wtr hdr splitted-files mode key-fn))))
+  (let [temp-dir (util/create-temp-dir "cljam.also.sorter")
```
alumi: `"cljam.algo.sorter"`

r6eve:
Thank you for reviewing and I apologize for my poor understanding.
I fixed them and squashed the commits into two commits.

r6eve:
Thank you for correcting me and merging!

PR #179, opened 2019-09-26
Avoid an ArityException on creating an index for BAM files without alignments

alumi:
#### Problem
`cljam.algo.bam-indexer/create-index` throws an `clojure.lang.ArityException` when the target BAM contains no alignments

#### Cause
An initial value is not provided for `(reduce merge-index [])` in `cljam.io.bam-index.writer/make-index-from-blocks`
https://github.com/chrovis/cljam/blob/2db7fe739047e2284e93e9d211be74cfdd6cc09b/src/cljam/io/bam_index/writer.clj#L310

#### Changes
Add a default value for `make-index-from-blocks`

#### Affects
- `cljam.algo.bam-indexer`
- `cljam.io.bam-index.writer`

#### Tests

- `lein check` ✅
- `lein test :all` ✅

### DIFF
--- a/src/cljam/io/bam_index/writer.clj
+++ b/src/cljam/io/bam_index/writer.clj
@@ -138,7 +138,7 @@
                        indices)]
         (recur rest rid' idx-status' no-coordinate-alns' indices'))
       (assoc indices rid idx-status
-                     :no-coordinate-alns no-coordinate-alns))))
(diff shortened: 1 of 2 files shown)


alumi:
Thank you for reviewing! @r6eve @yito88

PR #170, opened 2019-07-01
Provide a default file format version for BCF

alumi:
~~**This PR depends on #169**
The actual diff is [94ce1f8...581438a](https://github.com/chrovis/cljam/pull/170/files/94ce1f8de25130bc2a28aa5a03e3fec3e34672f9..581438a1502bf4ab1af0d4b1ea74970b3d876320)~~

Like VCFWriter, this PR provides a default file format version `"VCFv4.3"` for meta-info of BCF.
Though BCF has its version in the magic number `BCF\2\2`, the header string should start with `##fileformat=`.

PR #181, opened 2019-11-08
Migrate CI to GitHub Actions

r6eve:
Travis CI is sometimes slow to start and changes the provided JDK versions drastically https://github.com/chrovis/cljam/pull/170. To better achieve stability, it's time to switch CI from Travis CI to GitHub Actions.


.github/workflows/main.yml
```diff
@@ -0,0 +1,59 @@
...
+        key: ${{ runner.os }}-m2-${{ hashFiles('**/project.clj') }}
+        restore-keys: |
+          ${{ runner.os }}-m2-
+    - name: Install Dependencies
+      run: lein deps
+    - name: Run tests
+      run: |
+        lein with-profile +dev:+1.8:+1.9 test
+        lein with-profile dev bin
+        target/cljam version
```
r6eve: Duplicated code here. I tried to include EA in Java's metrix, and found it's impossible for now.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v1
+    - name: Setup Java
+      uses: actions/setup-java@v1
+      with:
+        java-version: '1.8'
+        architecture: x64
+    - name: Deploy
+      run: ./deploy-snapshot.sh
```
alumi: I want this workflow to be executed after all jobs in the `main` workflow have been finished successfully.
I think keeping workflows separated has advantages of maintainability and security but it seems we can't define dependencies for workflows for now.
Do we have to merge this job into `main` with [`if`](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idif) directive?

r6eve: Yes. As far as I know, there's no way except `if` directive.
I incorporated the `deploy` job into the `main` workflow with some constraints. When pushing master, if any of `main` tests fails, the deploy is not executed.

r6eve:
Thanks for reviewing.

PR #182, opened 2019-11-19
Add support for CSI reader.

niyarin:
This PR adds support for CSI reader to read BCF file randomly.
I've modify functions for other index file so that CSI can be handled by io/util/bin.
The referenced specification is CSIV1.


src/cljam/io/util/bin.clj
```diff
@@ -37,8 +40,18 @@
...
+        min-offsets (get-min-offset index-data ref-idx beg bins)]
+    (->>
+     (mapcat
+      (fn [chunks min-offset]
+        (filter
+         (fn [^Chunk x]
+           (> (.end x) min-offset))
+         chunks))
+      chunks-list
+      min-offsets)
```
alumi: I don't get the point of comparing min-offset to chunks in the same bin.
I think a min-offset of the smallest bin with the smallest index is sufficient for all chunks.

```
|=======================0=======================|
|-----------1-----------|===========2===========|
|-----3-----|-----4-----|=====5=====|=====6=====|
|--7--|--8--|--9--|--10-|--11-|==12=|==13=|==14=|
                           |-x-|
                    |---y---|
                    |-----y'-----|

                                   |====q====|
```

Let's consider a binary tree instead of an octree.
Given a query interval `q`, chunks in overlapping bins `[0, 2, 5, 6, 12, 13, 14]` will be candidates.
Let `x` (which itself belongs to the bin `5`) be an entity with the smallest starting position overlapping the bin `12`. 
Chunks ending before the starting file offset of `x` in the bins `[0, 2, 5, 6]` can be ignored since they never overlap the bin `12`. (e.g. `y` in the bin `0` can be ignored)
Entities like `y'` never exist otherwise the l-offset of the bin `12` would be `y'` instead of `x`.

We can apply this method recursively for each depth (use l-offset of `12` for filtering chunks in the bins `[0, 2, 5, 6]`, `5` for `[0, 2]`...), but I think it's unnecessary for now.

src/cljam/io/util/bin.clj
```diff
@@ -1,33 +1,36 @@
...
+       (let [t
+             (apply + (map (fn [x] (bit-shift-left 1 (* x 3))) (range (+ d 1))))
+             s
+             (+ min-shift
+                (* 3
+                   (- depth d 1)))]
+         (range (+
+                 t
+                 (bit-shift-right beg s))
+                (+ t 1 (bit-shift-right (- end 1) s)))))
```
alumi:
```suggestion
                (+ t 1 (bit-shift-right (dec end) s)))))
```

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,78 @@
...
+  "Reader of a CSI format file."
+  (:require [cljam.io.util.bgzf :as bgzf]
+            [cljam.io.util.lsb :as lsb]
+            [cljam.io.util.bin :as util-bin]
+            [clojure.string :as cstr])
+  (:import java.util.Arrays
+           [java.io DataInputStream IOException]
+           [cljam.io.util.chunk Chunk]))
+
+(deftype Csi [n-ref min-shift depth bidx loffset]
```
r6eve: `CSI` sounds good to me bacause CSI is an acronym that stands for Coordinate Sorted Index.
https://guide.clojure.style/#CamelCase-for-protocols-records-structs-and-types

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,78 @@
...
+          (comp
+           (map (get (.loffset this) ref-idx))
+           (filter identity))
+          bins))
+  (get-depth [this]
+    (.depth this))
+  (get-min-shift [this]
+    (.min-shift this)))
+
+(def csi-magic "CSI\1")
```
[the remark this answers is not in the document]
r6eve: Also, add `:private`.

src/cljam/io/util/bin.clj
```diff
@@ -37,8 +40,18 @@
 (defn get-spans
   "Calculate span information for random access from index data such as tabix."
```
r6eve:
```suggestion
   "Calculates span information for random access from index data such as tabix."
```
As cljam's convention, docstring uses third-person singular verbs. Please fix it if you notice old misspellings.

niyarin:
Thank you for pointing.
I fixed them.

src/cljam/io/util/bin.clj
```diff
@@ -1,33 +1,36 @@
...
+                   (- depth d 1)))]
+         (range (+
+                 t
+                 (bit-shift-right beg s))
+                (+ t 1 (bit-shift-right (- end 1) s)))))
+     (range depth))))
 
 (def ^:private reg->bins (memoize reg->bins*))
 
 (defprotocol IBinaryIndex
```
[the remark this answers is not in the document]
r6eve: I'm in favor of `IBinningIndex` since the word `binning index` is used in the specification.

src/cljam/io/util/bin.clj
```diff
@@ -1,44 +1,45 @@
 (ns cljam.io.util.bin
-  (:require [cljam.io.util.chunk :as util-chunk]))
+  (:require [cljam.io.util.chunk :as util-chunk])
+  (:import [cljam.io.util.chunk Chunk]))
```
r6eve: Unused imports here. Using [clj-kondo](https://github.com/borkdude/clj-kondo/blob/master/doc/editor-integration.md), we can catch that.

test/cljam/io/csi_test.clj
```diff
@@ -0,0 +1,30 @@
+(ns cljam.io.csi-test
+  (:require [clojure.java.io :as cio]
+            [clojure.test :refer :all]
+            [cljam.test-common :refer :all]
+            [cljam.io.csi :as csi])
+  (:import
+   [cljam.io.csi CSI]
+   [cljam.io.util.chunk Chunk]))
```
r6eve: Ditto.

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,76 @@
...
+    (into [] (mapcat (get bidx ref-idx) bins)))
+  (get-min-offset [_ ref-idx beg bins]
+    (let [oct-cumulative-sums
+          (->> (range (inc depth))
+               (map #(bit-shift-left 1 (* % 3)))
+               (reductions +))
+          min-index
+          (->> (filter #(< % (last bins)) oct-cumulative-sums)
+               last
+               inc)
```
r6eve: `oct-cumulative-sums` is sorted by ascending order, so `take-while` is better.
```clojure
(->> (take-while #(< % (last bins)) oct-cumulative-sums)          
     last          
     inc)
```

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,76 @@
...
+          target-bin
+          (first
+           (filter #(< min-index %) bins))
```
r6eve: `bins` is also sorted.
```clojure
target-bin (first (take-while #(< min-index %) bins))
```

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,76 @@
...
+          res (get (loffset ref-idx) target-bin)]
+      (if res res 0)))
```
r6eve: This could be `(get (loffset ref-idx) target-bin 0)`

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,76 @@
...
+            [cljam.io.util.bin :as util-bin])
+  (:import java.util.Arrays
+           [java.io DataInputStream IOException]
+           [cljam.io.util.chunk Chunk]))
+
+(deftype CSI [n-ref min-shift depth bidx loffset]
+  util-bin/IBinaryIndex
+  (get-chunks [_ ref-idx bins]
+    (into [] (mapcat (get bidx ref-idx) bins)))
+  (get-min-offset [_ ref-idx beg bins]
```
alumi: I'm sorry I forgot that CSI doesn't always store l-offsets for all bins like BAI.
So in this comment https://github.com/chrovis/cljam/pull/182#discussion_r348276378, when bin `12` doesn't exist in CSI, we have to check the left bin `11` and then the parent bin `5` ... recursively until we find an existing bin.
Note that the bin `11` is not contained in the result of `reg->bins` in this case.
The current implementation will return the l-offset of the bin `13` if `12` doesn't exist, which results in unintentionally filtering necessary chunks.

niyarin:
Thank you for pointing.
I fixed them.

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,72 @@
...
+    (let [oct-cumulative-sums (->> (range (inc depth))
+                                   (map #(bit-shift-left 1 (* % 3)))
+                                   (reductions +))
+          min-index (->> (take-while #(<= % (last bins)) oct-cumulative-sums)
+                         last)]
+      (or (->> bins
+               (drop-while #(>  min-index %))
```
r6eve: Extra spaces between `>` and `min-index`. And `partial` is preferable. https://guide.clojure.style/#partial
```suggestion
               (drop-while (partial > min-index))
```

alumi:
@niyarin Would you rebase this branch onto `master` (faa852458d1c99d3b893f7ec6914090df616d33c)? CI will fail in this branch. See #183 for more details.

niyarin:
Sorry for my late.
I fixed them.

src/cljam/io/util/bin.clj
```diff
@@ -5,39 +5,59 @@
...
+  (let [x (inc (quot (- 64 (Long/numberOfLeadingZeros bin)) 3))]
+    (cond-> x
+      (< bin (first-bin-of-level x)) dec)))
+
+(defn bin-beg ^long [^long bin ^long min-shift ^long depth]
+  (let [level (bin-level bin)]
+    (inc (* (- bin (first-bin-of-level level))
+            (bin-width-of-level level min-shift depth)))))
 
 (def ^:private reg->bins (memoize reg->bins*))
```
r6eve: The position of defining memoized version is a little too far from original version.

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,72 @@
...
+                               (into (sorted-map)
+                                     (comp (map (juxt :bin :loffset))
+                                           (filter #(<= (first %) max-bin))
+                                           (map
+                                            #(vector
+                                              (util-bin/bin-beg
+                                               (first %)
+                                               min-shift
+                                               depth)
+                                              (second %))))
```
r6eve: Prefer literal collection syntax https://guide.clojure.style/#literal-col-syntax

```clojure
(comp (map (juxt :bin :loffset))                                       
      (filter #(<= (first %) max-bin))                                       
      (map (fn [[bin loffset]]                                       
             [(util-bin/bin-beg bin                                       
                                min-shift                                       
                                depth)                                       
              loffset])))
```

test/cljam/io/util/bin_test.clj
```diff
@@ -3,12 +3,50 @@
...
+(deftest bin-beg-test
+  (doseq [index-shift [12 14 16]]
+    (let [min-size (bit-shift-left 1 index-shift)]
+      (is (=  1 (util-bin/bin-beg 4681 index-shift 5)))
+      (is (=  (+ 1 min-size) (util-bin/bin-beg 4682 index-shift 5)))
+      (is (=  (+ 1 (* min-size 2)) (util-bin/bin-beg 4683 index-shift 5)))
+      (is (=  1 (util-bin/bin-beg 585 14 5)))
+      (is (=  (+ 1 (* min-size 8)) (util-bin/bin-beg 586 index-shift 5)))
+      (is (=  (+ 1 (* min-size 16)) (util-bin/bin-beg 587 index-shift 5)))
+      (is (=  (+ 1 (* min-size 24)) (util-bin/bin-beg 588 index-shift 5))))))
```
r6eve: Extra spaces here.

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,72 @@
...
+  (get-depth [_]
+    depth)
+  (get-min-shift [_]
+    min-shift))
+
+(def ^:const ^:private csi-magic "CSI\1")
+
+(defn- read-chunks!
+  [rdr]
+  (->> #(Chunk. (lsb/read-long rdr) (lsb/read-long rdr))
```
r6eve: Prefer constructor functions. https://guide.clojure.style/#record-constructors

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,72 @@
...
+       (repeatedly (lsb/read-int rdr))
+       vec))
+
+(defn- read-bin-index
+  [rdr]
+  (->> #(hash-map
+         :bin (lsb/read-int rdr)
+         :loffset (lsb/read-long rdr)
+         :chunks (read-chunks! rdr))
```
r6eve: In current implementation of Clojure, function arguments are certainly evaluated from left to right, but it's not very well known style.
I think using let form before `hash-map` might be more reasonable for side-effects. Please correct me if I'm wrong.
https://clojure.org/reference/evaluation

niyarin:
Thank you for pointing out and suggestions.
I fixed them.

src/cljam/io/csi.clj
```diff
@@ -0,0 +1,73 @@
...
+  (get-depth [_]
+    depth)
+  (get-min-shift [_]
+    min-shift))
+
+(def ^:const ^:private csi-magic "CSI\1")
+
+(defn- read-chunks!
+  [rdr]
+  (->> #(chunk/->Chunk (lsb/read-long rdr) (lsb/read-long rdr))
```
r6eve: Sorry for nagging, would you use let form for side effects?

niyarin:
I'm sorry to overlook.
As a small fix, I added it to the previous commit.

niyarin:
Thank you for pointing.
I fixed them.

niyarin:
Thank you for reviewing! @r6eve @alumi
And I squashed the commits.

r6eve:
Thank you so much for your cooperation!

PR #183, opened 2019-12-02
Detect the latest build number of OpenJDK EA

alumi:
The build 14 of OpenJDK EA has been deleted after build 25 is published.
We need to get the latest build number automatically to adapt to future releases.
This PR picks up a tarball URL from the HTML of jdk.java.net then [`set-env`](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#set-an-environment-variable-set-env) it to use it from `actions/setup-java`.


alumi:
Thanks for reviewing! @athos @r6eve

PR #180, opened 2019-10-15
Add support for random reading of vcf file.

niyarin:
#### Summery
This PR adds support for random reading of bgzip compressed VCF files.
The tabix format was adjusted to bam index, and its reading function was integrated with bam-index to create io/util/bin.clj.

PR #184, opened 2019-12-11
Fix a bug in reading bgzipped VCF files

alumi:
## Problem
`cljam.io.vcf/read-variants` throws `java.lang.ClassCastException` for bgzipped VCF files.

#### Cause
#180 introduced switching of reader type between `java.io.BufferedReader` and `bgzf4j.BGZFInputStream` but did not modify the existing code.

#### Changes
Switch methods based on the type of reader instance.

#### Tests

- `lein check` ✅
- `lein test :all` ✅
- `lein cloverage` ✅


alumi:
@athos Thanks! I added an extra test case in 4ac7fe2.

alumi:
Thank you for reviewing @niyarin @athos!!

PR #185, opened 2020-01-08
Fix float reader for specific environments such as IcedTea.

niyarin:
In some environments such as "IcedTea",Float reader in Bcf reader doesn't  work as expected.
Specifically, the result of reading NaN is expected to be 0x800001, but the result will be different.
The fix is ​​to read as int and convert to float if not NaN.


niyarin:
Sorry,I missed warning.I'll fix it.

alumi:
Thanks!!

niyarin:
Thank you for reviewing.

PR #191, opened 2020-04-17
Apply cljfmt

alumi:
- Added a GitHub Action for `cljfmt check `
- Fixed current source codes to pass the check


alumi:
@athos @niyarin Thanks for your reviews!

PR #193, opened 2020-04-22
Update actions/checkout to v2

r6eve:
This PR bumps the version of actions/checkout to new major version.


r6eve:
Thank you for the review.

PR #194, opened 2020-04-24
Fix reading chrs when chrs in index file and chr are not equal.

niyarin:
- To read aux field of CSI and names field of tabix to get chr names of index file generated by htslib.

**Cause**
In CSI created by htslib, the chr index may not match the vcf header.
It is the case that chr exists in the header of vcf but does not exist in the actual data, or the order of chr is changed.

**Changes**
The chr names of CSI created by htslib is in the aux field.(Note that this aux field is not included in the specifications)
In tabix, chr name field is required.

1. If the CSI has an aux inserted by htslib, read it, otherwise read the VCF header.
2.  Read names field of Tabix.


niyarin:
Thank you for the patch.
I applied a patch and returned the protocol to its original location.

niyarin:
Thank you for the review.
I squashed the commits.

PR #198, opened 2020-05-29
Add base_sha option to clj-lint-action.

niyarin:
This PR adds the option of clj-lint-action.
Without this option, the second commit on the branch would not be able to take the diff.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,9 +7,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - uses: xcoo/clj-lint-action@v1.1.1
+    - name : master base sha
+      id : master_base_sha
+      if : github.ref  == 'refs/heads/master'
+      run: echo  "::set-output name=base_sha::${{ github.event.before }}"
+    - uses: xcoo/clj-lint-action@v1.1.5
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -12,4 +12,5 @@ jobs:
         linters: "[\"clj-kondo\" \"kibit\" \"eastwood\"]"
         github_token: ${{ secrets.GITHUB_TOKEN }}
         runner: ":leiningen"
+        base_sha: ${{ github.event.before }}
```
alumi, line 15: I'm sorry if I misunderstood, but I guess `github.event.pull_request.base.sha` is more suitable for PR events since files eventually changed in a PR is more important than files changed in the last commit of a PR.

niyarin:
Thank you for the pointing.
I changed the options for pull requests as well as push.

alumi:
Hmm.. 
In https://github.com/chrovis/cljam/pull/198/checks?check_run_id=719698533, the `github.event.before` on a push event seems to refer to 0444810e31ac39cca1318da84ba73369de8e63ff which is a force-pushed obsolete commit :thinking:

niyarin:
For a non-master branch, I changed the action to pick up the changes from the beginning of the branch.
・Changed clj-lint-action to take a diff from the head of that branch when passing an empty commit sha.
・Changed lint.yml to take base_sha only for master.

PR #200, opened 2020-06-03
Add support cloning VCF/BCF.

niyarin:
This PR adds a function to create a new object by copying except stream when VCF/BCF object is put in the reader.
This function is almost the same as "cljam.io.bam.core/clone-reader".

### DIFF
--- a/src/cljam/io/vcf.clj
+++ b/src/cljam/io/vcf.clj
@@ -65,7 +65,7 @@
     (BCFReader. (.url rdr) (.meta-info rdr) (.header rdr)
                 input-stream (.start-pos rdr) (.index-delay rdr))))
 
-(defn clone-reader
+(defn ^Closeable clone-reader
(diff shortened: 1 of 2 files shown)


src/cljam/io/vcf.clj
```diff
@@ -46,16 +46,46 @@
...
+
+(defn ^BCFReader clone-bcf-reader
+  "Clones bcf reader sharing persistent objects."
+  [^BCFReader rdr]
+  (let [url (.url rdr)
+        input-stream (bgzf/bgzf-input-stream url)]
+    (BCFReader. (.url rdr) (.meta-info rdr) (.header rdr)
+                input-stream (.start-pos rdr) (.index-delay rdr))))
+
+(defn clone-reader
```
alumi, line 68:
```suggestion
(defn ^Closeable clone-reader
```

niyarin:
Thanks for the review!

PR #204, opened 2020-07-08
Introduce GitHub Dependabot to keep dependencies up to date

r6eve:
This PR adds settings for GitHub Dependabot to keep GitHub Actions dependencies up to date.
Currently, GitHub Dependabot checks only GitHub Actions dependencies, not others in `project.clj`.


r6eve:
Thank you for reviewing.

PR #212, opened 2020-10-15
Add support for indexed? for vcf/bcf.

niyarin:
Since `indexed?` of vcf / bcf was not implemented, I added it.
The implementation is almost the same as the `indexed?` of fasta.


alumi (approved):
LGTM 👍

r6eve (approved):
LGTM. Thanks for the helpful implementation.

PR #226, opened 2021-04-08
Update actions/setup-java to v2

alumi:
Related to: #224 

The differences between v1 and v2 are listed in https://github.com/actions/setup-java#v2-vs-v1
- v2 requires `distribution:` value, which is set to `adopt` here
- v2 requires `java-version:` to be a SemVer


athos (approved):
LGTM 👍

xckitahara (approved):
LGTM:+1:

PR #239, opened 2021-06-16
Fix bam encoder/decoder to convert sam to bam.

### LINKED ISSUE #238: Ambiguity in empty SEQ representation of SAM/BAM  format
If reading an alignment with an "empty" (void) SEQ from a SAM file using the SAM reader, you'll get something like:

```
{..., :seq "*", ...}
```

while you'll get `{..., :seq "", ...}` if you read such an alignment from a BAM file using the BAM reader.

Similarly, the SAM/BAM writers expect `"*"` / `""`, respectively, as the empty SEQ representation.

So, users wouldn't get in trouble with this inconsistency as far as they use the same file format as their input/output. But once trying to convert SAM to BAM (or vice versa), they would see an error or unexpected behavior.

Considering that empty (void) values for most other fields are represented as `"*"`, I think it would be better to make the empty SEQ representation of BAM aligned with that of SAM.

niyarin:
This PR fixes #238 .

I changed the bam encoder so that seq is an empty string when an alignment that is * is entered in seq.
And I changed the bam decoder so that seq is * when an alignment that is empty string is entered in seq.

As a side note, if you convert sam with seq * to bam with samtools, the seq of that bam will be an empty string. Also, if you `samtools view` the empty string bam, seq will be output as *. I think this cljam change will be similar to this behavior.


src/cljam/io/bam/decoder.clj
```diff
@@ -124,7 +124,10 @@
          cigar-bytes (lsb/read-bytes buffer (* n-cigar-op 4))
          [cigar ^long len] (cigar/decode-cigar-and-ref-length cigar-bytes)
          ref-end     (int ^long (if (zero? len) pos (dec (+ pos len))))
-         seq         (decode-seq (lsb/read-bytes buffer (quot (inc l-seq) 2)) l-seq)
+         seq         (if (= l-seq 0)
```
alumi, line 127:
```suggestion
         seq         (if (zero? l-seq)
```

BTW, I thought the annotation at the top should actually point to this line since the hunk is contained in this PR. Would you take a look at `clj-lint-action`?

image

src/cljam/io/bam/decoder.clj
```diff
@@ -151,7 +154,9 @@
              [cigar ^long len] (cigar/decode-cigar-and-ref-length cigar-bytes)
              ref-end     (int ^long (if (zero? len) pos (dec (+ pos len))))]
          (when (<= start ref-end)
-           (let [seq     (decode-seq (lsb/read-bytes buffer (quot (inc l-seq) 2)) l-seq)
+           (let [seq     (if (= l-seq 0)
```
alumi, line 157:
```suggestion
           (let [seq     (if (zero? l-seq)
```

alumi (approved):
Thank you for the PR! LGTM 👍 
I added some trivial comments.

src/cljam/io/bam/decoder.clj
```diff
@@ -124,7 +124,10 @@
          cigar-bytes (lsb/read-bytes buffer (* n-cigar-op 4))
          [cigar ^long len] (cigar/decode-cigar-and-ref-length cigar-bytes)
          ref-end     (int ^long (if (zero? len) pos (dec (+ pos len))))
-         seq         (decode-seq (lsb/read-bytes buffer (quot (inc l-seq) 2)) l-seq)
+         seq         (if (zero? l-seq)
+                       "*"
+                       (decode-seq (lsb/read-bytes
+                                    buffer (quot (inc l-seq) 2)) l-seq))
```
athos, line 130: I’d prefer that `decode-seq` has the same level of granularity as `decode-qual`. If `decode-qual` returns `"*"` for an empty qual, users would expect `decode-seq` to return `"*"` for an empty seq.

With regard to code redundancy as well, I think it would be better to move the code dealing with `"*"` into `decode-seq` itself.

src/cljam/io/bam/encoder.clj
```diff
@@ -83,39 +83,42 @@
...
+    (lsb/write-bytes wrtr (seq/str->compressed-bases (:seq aln)))
+    ;; qual
+    (lsb/write-bytes wrtr (encode-qual aln))
+
+    ;; options
+    (doseq [op (:options aln)]
+      (let [[tag value] (first (seq op))]
+        (lsb/write-short
+         wrtr (short (bit-or (bit-shift-left (byte (second (name tag))) 8)
+                             (byte (first (name tag))))))
```
athos, line 122: This is somewhat nitpicky, but the indentation here looks a little bit tricky.
The Clojure style guide [says](https://github.com/bbatsov/clojure-style-guide#vertically-align-fn-args): "Vertically align function (macro) arguments spanning multiple lines."

So, I think:

```clojure
(lsb/write-short wrtr
                 (short (bit-or (bit-shift-left (byte (second (name tag))) 8)
                                (byte (first (name tag))))))
```

or:

```clojure
(lsb/write-short
 wrtr
 (short (bit-or (bit-shift-left (byte (second (name tag))) 8)
                (byte (first (name tag))))))
```

or even something like:

```clojure
(let [xxx (short (bit-or (bit-shift-left (byte (second (name tag))) 8)
                         (byte (first (name tag)))))]
  (lsb/write-short wrtr xxx))
```

looks more natural.

athos (commented):
Thanks for fixing the issue, @niyarin!

Basically, the change looks good 👍 I just left a few minor comments.

niyarin:
Thank you for the comment, @athos. I fixed them.

athos (approved):
Thanks for updating the code! LGTM 👍

PR #247, opened 2021-09-29
Fix error that occurs when querying intervals for missing chr

athos:
When issuing an interval query (using `c.u.i/find-overlap-intervals`) for a missing chr in the index, an error will be thrown:

```clojure
(require '[cljam.util.intervals :as intervals])

(def index
  (intervals/index-intervals
   [{:chr "chr1", :start 10000, :end 15000}
    {:chr "chr1", :start 20000, :end 25000}]))

(intervals/find-overlap-intervals index "chr1" 13000 17000)
;=> ({:chr "chr1", :start 10000, :end 15000})
(intervals/find-overlap-intervals index "chr1" 17000 19000)
;=> ()

(intervals/find-overlap-intervals index "chr2" 13000 17000)
;; Execution error (IllegalArgumentException) at cljam.util.intervals/eval5662$fn$G (intervals.clj:4).
;; No implementation of method: :find-overlap-intervals* of protocol: #'cljam.util.intervals/IIntervals found for class: nil
```

This PR fixes the error by guarding against `nil` with `some->`.

### DIFF
--- a/src/cljam/util/intervals.clj
+++ b/src/cljam/util/intervals.clj
@@ -64,4 +64,4 @@
   "Find intervals that are on the given `chr` and overlap the given interval
   [`start` `end`] using indexes created by `index-intervals`."
   [indexed-intervals chr start end]
-  (find-overlap-intervals* (get indexed-intervals chr) start end))
+  (some-> (get indexed-intervals chr) (find-overlap-intervals* start end)))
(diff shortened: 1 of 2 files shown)


alumi (approved):
Wow, thank you for the fix! :+1:

niyarin (approved):
Thank you for fixing. LGTM :+1: 
(I wrote the original code , but I overlooked the case with no corresponding chr.)

athos:
@alumi @niyarin Thanks for the quick review!

@alumi Do you think it's worth cutting another release soon?

alumi:
Sure! I'll release `0.8.2` in an hour 👍

athos:
Thanks a lot 🙏 I'm looking forward to it!

PR #253, opened 2021-12-24
Use bgzip as default for writing `.gz` files

### LINKED ISSUE #251: Bgziped VCF is illegal.
We can reproduce it with the following code.
```clojure
(with-open [wtr (vcf/writer "./test.vcf.gz"
                            {:format nil :contig {:id "1", :length nil} :filter [{:id "PASS", :description "All filters passed"}] :info nil}
                            ["CHROM" "POS" "ID" "REF" "ALT" "QUAL" "FILTER" "INFO" "FORMAT"])]
  (vcf/write-variants wtr [{:chr "1" :pos 10 :ref "A" :alt ["G"]}]))

(with-open [rdr (vcf/reader "./test.vcf.gz")]
  (vcf/read-variants rdr))
;Error printing return value (IOException) at java.io.BufferedReader/ensureOpen (BufferedReader.java:122).
```
The generated bgzip file also fails with the'bcftools index'.
Now, if we do `bgzip -d` and` bgzip`, the error doesn't occur.

alumi:
Fixes #251 

This PR changes the behavior of `cljam.util/compressor-output-stream`.
It now returns `BGZFOutputStream` instead of `GzipCompressorOutputStream` for files with `.gz` extension.
Since BGZIP is just a special case of GZIP, existing codes using `cljam.util/compressor-output-stream` with `.gz` should still work.
To get `GzipCompressorOutputStream`, we can use the 2nd argument or `.gzip` extension.

This PR also includes
- Update log4j to 2.17.0
- Some minor fixes suggested by linters

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -35,7 +35,7 @@
              :1.9 {:dependencies [[org.clojure/clojure "1.9.0"]]}
              :1.10 {:dependencies [[org.clojure/clojure "1.10.3"]]}
              :uberjar {:dependencies [[org.clojure/clojure "1.10.3"]
-                                      [org.apache.logging.log4j/log4j-api "2.14.1"]
+                                      [org.apache.logging.log4j/log4j-api "2.17.0"]
                                       [org.apache.logging.log4j/log4j-core "2.14.1"]]
                        :resource-paths ["bin-resources"]
                        :main cljam.tools.main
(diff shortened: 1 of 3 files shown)


r6eve (approved):
LGTM

niyarin (approved):
LGTM :+1:

alumi:
@r6eve @niyarin Thank you for reviewing! 👍

PR #264, opened 2022-09-05
Fix error that occurs when encoding a value greater than 32767 with  ‘B, S’ data type in encoding alignment in encoding alignment.

matsutomo81:
In encoding alignment options, encoding values greater than 32767 with the 'B, S' data type causes an error.

For example, encoding the following alignment would result in the following error message. Note that the alignment option has the value of '65535' for 'B, S' type.
`
q006	0	ref	10	60	16M	*	0	0	AAAAGGGGTTTTCCCC	*	XS:B:S,0,32767,65535
`

```
; Execution error (NumberFormatException) at java.lang.Short/parseShort (Short.java:139).
; Value out of range. Value:"65535" Radix:10
```
The reason for error is as follows. 
According to [BAM format specification](https://github.com/samtools/hts-specs/blob/master/SAMv1.tex#L1175), the value of 'B, S' type are encoded with `uint16_t`, but the current code was processing with `Short/parseShort`. Therefore, when encoding values greater than 32767, `Short/parseShort` returns an error.

This PR solves this problem by removing `Short/parseShort`and using `lsb/write-ushort`. This PR also adds tests to convert SAM/BAM file that contains the alignments with 'B, S' type data.


alumi (changes requested):
Thank you for fixing this!
It seems that the BAM file you committed (`test-resources/bam/include-BS-type-option.bam`) doesn't contain `B:S` options. Would you replace it with a one generated by `samtools view -bh test-resources/sam/include-BS-type-option.sam`?

matsutomo81:
@alumi 
Thank you for finding my mistakes!
The data in the test BAM file was incorrect and has been fixed.

alumi (approved):
LGTM 👍 Thanks!

PR #265, opened 2022-09-09
Extend the supported data types in encoding alignment options

matsutomo81:
The topic of this PR is about encoding SAM format into BAM format. According to [BAM format specification](https://github.com/samtools/hts-specs/blob/59a0d0c8dc01b46192f6da1cec61db59c9b3d1d9/SAMv1.tex#L1117), there are several data types for options in the alignment, but [current cljam doesn't support all of them for encoding](https://github.com/chrovis/cljam/blob/c7cd8c43f13432d1cbcc95640108f7d32c6dcdef/src/cljam/io/bam/encoder.clj#L52). If you try to encode a data type that is not supported, an Execution error will be returned.

This PR extends the data types of options that can be encoded, so that all data types except the H-type can be supported. Since the H-type seems to be a specification that remains only for backward compatibility, this PR did not provide support for the H-type encoding. Tests have also been added for the functions I have modified.


test/cljam/io/bam/encoder_test.clj
```diff
@@ -0,0 +1,64 @@
...
+            [cljam.io.util.lsb :as lsb]
+            [cljam.io.bam.encoder :as encoder]
+            [cljam.io.bam.decoder :as decoder]))
+
+(deftest encode-tag-value-except-B-tag
+  (are [val-type value]
+       (= value
+          (as-> (#'encoder/encode-tag-value (lsb/gen-byte-buffer 100) val-type value) bb
+            (.position bb 0)
+            (decoder/parse-tag-single val-type bb)))
```
matsutomo81, line 14: The test is not completed only by the encoder, but also depends on the decoder. Should I rewrite it?

test/cljam/io/bam/encoder_test.clj
```diff
@@ -0,0 +1,64 @@
...
+    \f (Float/POSITIVE_INFINITY)
+    \Z "aaaBBB0011223344@@@+++"
+    \Z "!\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~"))
+
+(deftest encode-tag-value-only-B-tag
+  (are [value]
+       (= value
+          (as-> (#'encoder/encode-tag-value (lsb/gen-byte-buffer 100) \B value) bb
+            (.position bb 0)
+            (#'decoder/parse-tag-array bb)))
```
matsutomo81, line 49: It depends on the decoder as well as the above comment.

test/cljam/io/bam/encoder_test.clj
alumi: I agree that this kinds of tests are useful but they should not be here. Would you move them to somewhere like `cljam.io.bam-test`?
This namespace should have tests depending only on `cljam.io.bam.encoder` and its dependencies, like the following:

```clj
(deftest encode-tag-value-test
  (testing "non-array types"
    ;; ...
    )
  (testing "NULL-terminated text"
    ;; ...
    )
  (testing "array types"
    (are [?tag ?values ?bytes]
        (= (map unchecked-byte ?bytes)
           (let [bb (ByteBuffer/allocate 100)]
             (#'encoder/encode-tag-value
              bb \B (string/join \, (cons ?tag ?values)))
             (take (.position bb) (.array bb))))
      \c [0 127 -128 -1]
      [0x63                ;; c
       0x04 0x00 0x00 0x00 ;; 4
       0x00 0x7f 0x80 0xff]

      \C [0 127 128 255]
      [0x43                ;; C
       0x04 0x00 0x00 0x00 ;; 4
       0x00 0x7f 0x80 0xff]

      ;; ...
      )))
```

alumi (changes requested):
Thank you for another PR @matsutomo81 👍 I added [a comment](https://github.com/chrovis/cljam/pull/265#discussion_r969311361) about your concerns.
Since I just merged #264, please rebase and force-push to this branch when you finish editing the tests 🙏

test/cljam/io/bam/encoder_test.clj
```diff
@@ -0,0 +1,107 @@
...
+      \f -17.25 [0x00 0x00 0x8a 0xc1]
+      \f 0.0 [0x00 0x00 0x00 0x00]
+      \f 5.75 [0x00 0x00 0xb8 0x40]
+      \f (Float/MAX_VALUE) [0xff 0xff 0x7f 0x7f]
+      \f (Float/MIN_NORMAL) [0x00 0x00 0x80 0x00]
+      \f (Float/MIN_VALUE) [0x01 0x00 0x00 0x00]
+      \f (Float/NEGATIVE_INFINITY) [0x00 0x00 0x80 0xff]
+      \f (Float/POSITIVE_INFINITY) [0x00 0x00 0x80 0x7f]))
+
+  (testing "NULL-terminated text"
```
matsutomo81, line 51: I thought this part should use the new `testing` to update the description. As a result, the content of `are` in this `testing` is exactly the same as the previous one. 
The following `are` is almost the same as well.
https://github.com/chrovis/cljam/blob/c47ebe395848637db69070fb0580323b49e69fb4/test/cljam/io/bam/encoder_test.clj#L63

Do you think it is fine as it is? Or do you have any advice on how I should write it better?

alumi (changes requested):
When you are done editing a PR, please click the ["Re-request review" button on the reviewers list](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review#:~:re-review) 🙏

alumi (approved):
Thank you for updating the patch. Looks good 👍 
It is recommended not to force-push to the PR branch while being reviewed on GitHub. This is meant for keeping track of inline review comments across commits. You can clean your commits up when all reviewers approve your PR if necessary.

xckitahara (approved):
LGTM 👍

PR #266, opened 2022-11-16
Fix cljam.io.bam-test to work properly

matsutomo81:
This PR resolves two issues included in the current `cljam.io.bam-test`.
#### First problem:
Current `cljam.io.bam-test` fails when run with JDK 19. This PR will fix `cljam.io.bam-test` to work regardless of JDK version.
Accordiing to [this release note](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8291475) related to JDK 19, the result of `Float.toString(float)` evaluation may change from previous releases. The value of `1.17549435E-38`, which is used in current test, is affected by this and It causes test failures. This PR resolves the problem by avoiding the use of `1.17549435E-38`.
#### Second problem:
In the following code, there is a `?value` at the end of the `as->` macro, thus effectively evaluating `(= ?value ?value)`. Perhaps a previous typo, which I have removed.
https://github.com/chrovis/cljam/blob/843e30b4aaf34e2fe79d070f1a2122097eb78b99/test/cljam/io/bam_test.clj#L10-L15
Same as above.
https://github.com/chrovis/cljam/blob/843e30b4aaf34e2fe79d070f1a2122097eb78b99/test/cljam/io/bam_test.clj#L53-L58


matsutomo81:
By referring to the log below, you can see that the current test fails on Java 19.
https://github.com/chrovis/cljam/actions/runs/3485121567/jobs/5830364082#step:7:305

alumi (changes requested):
As I commented before, `(= ?value ?value)` doesn't check anything.
Fix the assertion first, then update the test cases. The linter warnings will be also resolved.

alumi:
Uh, I see. `(= ?value ?value)` doesn't concern array types. Anyway, you can fix it in this PR since the goal of making the `cljam.io.bam-test` work properly is the same 👍

.github/workflows/main.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        java: [ '8', '11', '16', '17-ea']
+        java: [ '8', '11', '16', '17-ea', '19']
```
alumi, line 10: LTS versions and the latest

```suggestion
        java: [ '8', '11', '17', '19']
```

alumi (approved):
LGTM 👍

alumi:
I just changed the setting of reviewers assignment from load balancing to round robin and redrew a reviewer.

niyarin (approved):
LGTM. 
Sorry for the late review.

PR #269, opened 2023-04-24
Fix/apply lint

niyarin:
I fixed  almost bad coding style, except Boxed-math, which I couldn't fix.
#### master
``` sh
> find src/ |grep clj|xargs clj-kondo --lint
linting took 1340ms, errors: 1, warnings: 85
> find test/ |grep clj|xargs clj-kondo --lint
linting took 2115ms, errors: 1, warnings: 3
```
#### This PR
```sh
> find src/ |grep clj|xargs clj-kondo --lint
linting took 1325ms, errors: 0, warnings: 0
> find test/ |grep clj|xargs clj-kondo --lint
test/cljam/io/gff_test.clj:319:70: error: Unresolved symbol: ?str
linting took 2240ms, errors: 1, warnings: 0
```

#### Master
``` sh
> lein eastwood "{:linters [:all] :exclude-linters [:keyword-typos :unused-namespaces :boxed-math] }"
== Warnings: 11. Exceptions thrown: 0
```
#### This PR
``` sh
> lein eastwood "{:linters [:all] :exclude-linters [:keyword-typos :unused-namespaces :boxed-math] }"
== Warnings: 0. Exceptions thrown: 0
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,7 +10,7 @@ jobs:
     - name : master base sha
       id : master_base_sha
       if : github.ref  == 'refs/heads/master'
-      run: echo  "::set-output name=base_sha::${{ github.event.before }}"
+      run: echo "base_sha=${{ github.event.before }}" >> $GITHUB_OUTPUT
     - uses: xcoo/clj-lint-action@v1.1.11
       with:
         linters: "[\"clj-kondo\" \"kibit\" \"eastwood\"]"
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,3 +18,4 @@ jobs:
         runner: ":leiningen"
         base_sha: ${{ github.event.pull_request.base.sha||steps.master_base_sha.outputs.base_sha }}
         eastwood_linters: "[:all]"
+        linter_options: "{:eastwood {:exclude-linters [:keyword-typos :unused-namespaces :boxed-math :reflection]}}"
--- a/src/cljam/io/bam_index/writer.clj
+++ b/src/cljam/io/bam_index/writer.clj
@@ -96,8 +96,8 @@
         win-end (if (zero? aln-end)
                   win-beg
                   (util-bin/pos->lidx-offset aln-end linear-index-shift))
-        min* (fn [x]
-               (if x (min x beg) beg))]
+        min* (fn ^long [x]
+               (if x (min (long x) beg) beg))]
     (loop [i win-beg, ret linear-index]
       (if (<= i win-end)
         (recur (inc i) (assoc ret i (min* (get ret i))))
@@ -159,7 +159,7 @@
   "Complements a linear index.
   e.g. ([1 10] [3 30]) -> ([0 0] [1 10] [2 10] [3 30])"
   [linear-index]
-  (loop [[f & r] (if (zero? (ffirst linear-index))
+  (loop [[f & r] (if (zero? (long (ffirst linear-index)))
                    linear-index
                    (conj linear-index [0 0]))
          ret []]
@@ -210,7 +210,8 @@
 (defn merge-index
   "Merges two intermediate indices, returning the merged intermediate index."
   [idx1 idx2]
-  (let [no-coordinate-alns (+ (:no-coordinate-alns idx1) (:no-coordinate-alns idx2))
+  (let [no-coordinate-alns (+ (long (:no-coordinate-alns idx1))
+                              (long (:no-coordinate-alns idx2)))
         idx1 (dissoc idx1 :no-coordinate-alns)
         idx2 (dissoc idx2 :no-coordinate-alns)]
     (-> (merge-with
@@ -289,7 +290,7 @@
   "Update the last pointer of the index to the given value."
   [index eof-ptr]
   (if (or (= (keys index) [:no-coordinate-alns])
-          (pos? (get index :no-coordinate-alns 0)))
+          (pos? (long (get index :no-coordinate-alns 0))))
     index
     (let [last-ref (apply max (keys (dissoc index :no-coordinate-alns)))
           last-key (->> (for [[bin chunks] (get-in index [last-ref :bin-index])
--- a/src/cljam/io/bam_index/writer.clj
+++ b/src/cljam/io/bam_index/writer.clj
@@ -19,7 +19,7 @@
 
 (deftype BAIWriter [^DataOutputStream writer refs url]
   Closeable
-  (close [this]
+  (close [_]
     (.close writer)))
 
 
--- a/src/cljam/io/sam/util.clj
+++ b/src/cljam/io/sam/util.clj
@@ -16,7 +16,7 @@
   [line]
   (let [[qname flag rname pos-str mapq cigar rnext pnext tlen seq qual & options] (cstr/split line #"\t")
         pos (Integer/parseInt pos-str)
-        ref-length (cigar/count-ref cigar)
+        ref-length (int (cigar/count-ref cigar))
         end (if (zero? ref-length) 0 (int (dec (+ pos ref-length))))]
     (SAMAlignment. qname (Integer/parseInt flag) rname pos end (Integer/parseInt mapq)
                    cigar rnext (Integer/parseInt pnext) (Integer/parseInt tlen) (cstr/upper-case seq)
@@ -35,11 +35,12 @@
 
 (defn get-end
   "Returns the end position in reference for the given alignment."
-  [aln]
-  (let [ref-length (cigar/count-ref (:cigar aln))]
-    (if (zero? ref-length)
-      (:pos aln)
-      (dec (+ (:pos aln) ref-length)))))
+  ^long
+  [{:keys [^long pos cigar]}]
+  (let [ref-length (cigar/count-ref cigar)]
+    (if (zero? (long ref-length))
+      pos
+      (dec (+ pos (long ref-length))))))
 
 (defn compute-bin
   "Returns indexing bin based on alignment start and end."
--- a/src/cljam/io/sam/util/sequence.clj
+++ b/src/cljam/io/sam/util/sequence.clj
@@ -35,15 +35,14 @@
         out-bb (ByteBuffer/allocate result-len)]
     (dotimes [_ result-len]
       (let [u (.get in-bb)
-            l (byte (if (.hasRemaining in-bb) (.get in-bb) \=))]
+            l (byte (if (.hasRemaining in-bb) (.get in-bb) (byte (int \=))))]
         (->> (bit-and 0x7F l)
              (bit-or (bit-shift-left (bit-and 0x7F u) 7))
              (aget ^bytes two-bytes-to-compressed-bases-table)
              (.put out-bb))))
     (.array out-bb)))
 
-(def ^:const ^:private ^String
-  compressed-bases-to-bases-table
+(def ^:const ^:private compressed-bases-to-bases-table
   ;; Index: compressed base n containing two nibbles => 2n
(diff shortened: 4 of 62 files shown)


alumi (changes requested):
Thank you for working on this, @niyarin!  I've added some comments.

I would like to take this chance to resolve boxed maths, too.
After playing around with the codes, I found that there are several varieties of this problem.
The biggest difference is that the warning is raised by clojure compiler or by eastwood. It seems like that there are cases that the compiler handles primitive types properly but eastwood can't recognize the types.
For example, [`cljam.io.util.bgzf/get-block-address`](https://github.com/chrovis/cljam/blob/38b7918467a7090fa3dce8b501447ac5d2c5453c/src/cljam/io/util/bgzf.clj#L74) is reported as boxed math by `lein eastwood '{:linters [:boxed-math] :namespaces [cljam.io.util.bgzf]}'` but not by `lein update-in :global-vars assoc '*unchecked-math*' :warn-on-boxed -- check`.
This is because of `^:const` metadata attached to the vars. The compiler perform [inlining of such vars](https://github.com/clojure/clojure/blob/4090f405466ea90bbaf3addbe41f0a6acb164dbb/src/jvm/clojure/lang/Compiler.java#L7325) to replace them with its values which are primitive numeric literals, obviously typed.
```clj
(set! *unchecked-math* :warn-on-boxed)
(def x 1)
(+ x 1)
;; Boxed math warning, call: public static java.lang.Number clojure.lang.Numbers.unchecked_add(java.lang.Object,long).
(def ^:const x 1)
(+ x 1)
```
So this is a bug in the linter and there is no practical problem at all, and they will be compiled and executed as a primitive-type arithmetic operation. This kind of reports can be ignored.

For cases that affect the actual bytecode, let's use this PR to correct the problem, checking with [`lein check`](https://github.com/chrovis/cljam/commit/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R32-R34) and not eastwood.

First, we will address function definitions as follows.
For functions that always return a non-null primitive value, prefix the return primitive hint with the parameter vector. (e.g.  [`cljam.common/get-exec-n-threads`](https://github.com/chrovis/cljam/commit/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b#diff-37d821d4efe3d41908dbd6c939593c0d81a1a908bfab3350a372ac019c21ab46R13)) This will make the return type primitive on direct caller via var.

Also, attach a primitive hint to a parameter that is always a primitive value. Note that the parameter symbol need to be top-level and must not be destructured. (e.g. [`cljam.io.bam-index.reader/read-bin-index*!`](https://github.com/chrovis/cljam/commit/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b#diff-06014a024886a1ce8462414efdfeec4463a451e2482400181302043de12cddbbR68)) This makes the type of parameters in function definitions primitive, and also allows function calls directly via var to be made with primitive arguments.

A reference to a dynamic var that is bound to a primitive value is not resolved to a primitive type as is, so we need to cast it. (e.g. [`cljam.common/*n-threads*`](https://github.com/chrovis/cljam/blob/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b/src/cljam/common.clj#L14)) Since other than long/double can be used for casts, select the appropriate size.
A var bound to a primitive value that can be `^:const` should be marked as `^:const`. (e.g. [`cljam.io.bam.encoder/fixed-tag-size`](https://github.com/chrovis/cljam/commit/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b#diff-163324ebbbd01639d63ad8d8aef9298286de15398426032ee6ba166ee1201409R13-R14))
Non-primitive values that are `^:const` do not need a type hint. (e.g. [`cljam.io.csi/csi-magic`](https://github.com/chrovis/cljam/commit/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b#diff-4b1e73eb17f46b1343bc466668f8e81ad335b922e43ad4ed3b8d89bb5790ce95L12-R12))

For destructuring of parameters, local bindings, and partial expressions, avoid primitive hints and cast instead. (e.g. [`cljam.io.bcf.reader/reader`](https://github.com/chrovis/cljam/commit/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b#diff-40265c1eddd881dc5cbb5df284330e8145daf2a98ca39adf43947895414a997aL89-R89))

However, the cast from `Character` to `byte`/`short` fails when `*unchecked-math*` is not `nil`, so cast it to`int` before casting to `byte`/`short`. (e.g. [`cljam.io.pileup/upper-table`](https://github.com/chrovis/cljam/commit/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b#diff-64774fb77a39e37bce11675f355b6c22bbb75a333bb414c39f95f8a9624996a5R42-R53))

```clj
(set! *unchecked-math* nil)
(byte \A) ;; => 65
(short \A) ;; => 65
(int \A) ;; => 65
(long \A) ;; => 65
(set! *unchecked-math* :warn-on-boxed)
(byte \A)
;; Execution error (ClassCastException)
;; class java.lang.Character cannot be cast to class java.lang.Number (java.lang.Character and java.lang.Number are in module java.base of loader 'bootstrap')
(short \A)
;; Execution error (ClassCastException)
;; class java.lang.Character cannot be cast to class java.lang.Number (java.lang.Character and java.lang.Number are in module java.base of loader 'bootstrap')
(int \A) ;; => 65
(long \A) ;; => 65
```

I have pushed an incomplete commit to [`spike/boxed-math`](https://github.com/chrovis/cljam/compare/fix/apply-lint...spike/boxed-math) and hope it will be helpful.

src/cljam/io/sam/util.clj
```diff
@@ -15,8 +15,8 @@
   "Parse an alignment line, returning a map of the alignment."
   [line]
   (let [[qname flag rname pos-str mapq cigar rnext pnext tlen seq qual & options] (cstr/split line #"\t")
-        pos (Integer/parseInt pos-str)
-        ref-length (cigar/count-ref cigar)
+        pos (int (Integer/parseInt pos-str))
```
alumi, line 18:
```suggestion
        pos (Integer/parseInt pos-str)
```

src/cljam/io/sam/util/sequence.clj
```diff
@@ -35,14 +35,15 @@
         out-bb (ByteBuffer/allocate result-len)]
     (dotimes [_ result-len]
       (let [u (.get in-bb)
-            l (byte (if (.hasRemaining in-bb) (.get in-bb) \=))]
+            l (byte (if (.hasRemaining in-bb) (.get in-bb) (int \=)))]
```
alumi, line 38:
```suggestion
            l (if (.hasRemaining in-bb) (.get in-bb) (byte (int \=)))]
```

src/cljam/io/sam/util/sequence.clj
```diff
@@ -35,14 +35,15 @@
...
         (->> (bit-and 0x7F l)
              (bit-or (bit-shift-left (bit-and 0x7F u) 7))
              (aget ^bytes two-bytes-to-compressed-bases-table)
              (.put out-bb))))
     (.array out-bb)))
 
-(def ^:const ^:private compressed-bases-to-bases-table
+(def ^:const ^:private ^String
+  compressed-bases-to-bases-table
```
alumi, line 46:
```suggestion
(def ^:const ^:private compressed-bases-to-bases-table
```

src/cljam/tools/cli.clj
```diff
@@ -26,7 +26,7 @@
 (defn- exit
   "Exits the program with the status after printing the message."
   [status message]
-  (binding [*out* (if (zero? status) *out* *err*)]
+  (binding [*out* (if (zero? (int status)) *out* *err*)]
```
alumi, line 29:
```suggestion
  [^long status message]
  (binding [*out* (if (zero? status) *out* *err*)]
```

.github/workflows/lint.yml
```diff
@@ -10,11 +10,12 @@ jobs:
...
-      run: echo  "::set-output name=base_sha::${{ github.event.before }}"
+      run: echo "base_sha=${{ github.event.before }}" >> $GITHUB_OUTPUT
     - uses: xcoo/clj-lint-action@v1.1.11
       with:
         linters: "[\"clj-kondo\" \"kibit\" \"eastwood\"]"
         github_token: ${{ secrets.GITHUB_TOKEN }}
         runner: ":leiningen"
         base_sha: ${{ github.event.pull_request.base.sha||steps.master_base_sha.outputs.base_sha }}
         eastwood_linters: "[:all]"
+        linter_options: "{:eastwood {:exclude-linters [:keyword-typos :unused-namespaces]}}"
```
alumi, line 21: Let's use compiler warnings instead of eastwood

```suggestion
        linter_options: "{:eastwood {:exclude-linters [:keyword-typos :unused-namespaces :boxed-math]}}"
```

https://github.com/chrovis/cljam/blob/7eeaf02b9ef4167ac2104df01dd94364ffcfcb3b/.github/workflows/main.yml#L32-L35

alumi (changes requested):
@niyarin Thank you for updating the PR! I added a few extra comments, but basically the changes seem to be fine👍.

src/cljam/io/bam_index/writer.clj
```diff
@@ -96,8 +96,8 @@
         win-end (if (zero? aln-end)
                   win-beg
                   (util-bin/pos->lidx-offset aln-end linear-index-shift))
-        min* (fn [x]
-               (if x (min x beg) beg))]
+        min* (fn ^long [x]
+               (if x (min (long x) beg) beg))]
```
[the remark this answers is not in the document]
niyarin: In some cases, this x will be nil, so I think the original is probably good.

niyarin:
@alumi 
Thank you for pointing.
I fixed them.

niyarin:
I'll add `:reflectoin` to eastwood exclude linters to avoid this warnings.
```
Warning: [eastwood][reflection] call to method charAt can't be resolved.
Warning: [eastwood][reflection] call to method put on java.nio.CharBuffer can't be resolved (argument types: unknown).
Warning: [eastwood][reflection] call to method charAt can't be resolved.
Warning: [eastwood][reflection] call to method put on java.nio.CharBuffer can't be resolved (argument types: unknown).
```

src/cljam/io/bam_index/writer.clj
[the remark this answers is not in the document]
alumi: It's my fault. Thanks!

alumi (commented):
Thank you for updating. There seems to be several points left that need to be fixed.
- https://github.com/chrovis/cljam/pull/269#discussion_r1199870017
- https://github.com/chrovis/cljam/pull/269#discussion_r1177887113
- https://github.com/chrovis/cljam/pull/269#discussion_r1177881453

niyarin:
@alumi
I fixed what was not fixed.

alumi (approved):
Thank you @niyarin! LGTM 👍 
Would you squash your commits as follows before merging?
1. Replace `::set-output` with `$GITHUB_OUTPUT`
2. Change linter options
3. Fix return type hints
4. Fix boxed math warnings

alumi:
@niyarin Additionally, if you have time, would you consider whether some [optional (default level is `:off`) linters of `clj-kondo`](https://github.com/clj-kondo/clj-kondo/blob/master/doc/linters.md#warn-on-reflection) could be enabled?
For example, `:missing-docstring`, `:single-key-in`, `:shadowed-var`, and `:unused-value` seem worth enabling.
If there is a new warning, I would appreciate it if you could correct it in another PR.

PR #272, opened 2023-05-30
Fix/vcf float case insensitive

niyarin:


niyarin:
benchmark code
```clj
(ns cljam.io.vcf-bench
  (:require [libra.bench :refer [defbench are]]
            [libra.criterium :as c]
            [cljam.test-common :as tcommon]
            [cljam.io.vcf :as vcf])
  (:import cljam.io.vcf.reader.VCFReader))

(defn make-float-hack-meta [m]
  (update m
           :format
          #(map
             (fn [fmt]
               (if (= (:type fmt) "Integer")
                 (assoc fmt :type "Float")
                 fmt)) %)))

(def ^:private temp-file (str tcommon/temp-dir "/test.vcf"))

(defbench read-variants-long-bench
  (tcommon/prepare-cavia!)
    (tcommon/with-before-after {:before (tcommon/prepare-cache!)
                        :after (tcommon/clean-cache!)}
    (with-open [rdr (vcf/reader tcommon/test-large-vcf-file)
                wtr (vcf/vcf-writer temp-file (make-float-hack-meta (.meta-info rdr)) (.header rdr))]
      (let [samples (map keyword (drop 10 (.header rdr)))]
      (vcf/write-variants wtr
                          (map (fn [v]
                                 (reduce
                                   (fn [v k]
                                     (assoc-in v [k :DP] ##NaN))
                                   v
                                   samples))
                                (take 20000 (vcf/read-variants rdr))))))

    (are [f]
         (c/quick-bench
          (with-open [rdr (vcf/reader f)
                      r2 (VCFReader. (.url rdr) (make-float-hack-meta (.meta-info rdr))
                                     (.header rdr) (.reader rdr) (.index-delay rdr))]
            (doseq [x (take 30000 (vcf/read-variants r2))]
              (:pos x))))
      tcommon/test-large-vcf-file
      temp-file)))
```

niyarin:
I've tried methods other than regular expressions
```
(defn- parse-float [s]
    (let [have-sign (or (= (first s) \+) (= (first s) \-))
          fc (if have-sign (second s) (first s))]
      (if (Character/isLetter fc)
        (case fc
          (\I \i) ##Inf
          (\N \n) ##NaN
          (Float/parseFloat s))
        (Float/parseFloat s))))
```

src/cljam/io/vcf/util.clj
```diff
@@ -10,13 +10,22 @@
        (and (= 1 (.length ~s))
             (= \. (.charAt ~s 0)))))
 
+(defn- parse-float [s]
+  (let [m (re-matches
+           #"[+-]?(?i)(nan|infinity|inf)" s)]
+    (if m
+      (case (first (second m))
+        (\n \N) Float/NaN
+        (\i \I) Float/POSITIVE_INFINITY)
```
athos, line 19: Should `-inf` and `-infinity` return `Float/NEGATIVE_INFINITY`?

niyarin: You are right.
Thank you.

athos (commented):
Thanks for tackling this fix!

I took some profiles before and after the change, and confirmed that the overhead of `parse-float` got to 6% after the change, which feels somewhat noticeable compared to the overhead of 1% with the original code:

| Profile result before the change | Profile result after the change |
| :---: | :---: |
| Profile result before the change | Profile result after the change |

Digging into the details of the profile results, it turned out that the allocation of vectors to hold the submatch results (which occurs in `re-group` called in `re-matches`) and accessing those vectors accounts for the majority of the overhead.

Instead of `re-matches`, the direct use of `re-matcher` and the underlying `.matches`/`.group` methods reduces this overhead related to vector allocation, as follows. `parse-float` will be one of the most frequently called functions during VCF parsing, so you might want to pay much attention to allocation in it.

```clojure
(defn- parse-float [^String s]
  (let [m (re-matcher #"(?i)[+-]?(nan|inf(?:inity)?)" s)]
    (if (.matches m)
      (case (.charAt (.group m 1) 0)
        (\n \N) Float/NaN
        (\i \I) (if (= (.charAt s 0) \-)
                  Float/NEGATIVE_INFINITY
                  Float/POSITIVE_INFINITY))
      (Float/parseFloat s))))
```

- Profile result with the above implementation
Screenshot 2023-06-06 at 17 51 15

niyarin:
@athos 
Thank you for your profile.
I put your parse-float in the source code.

athos (approved):
Thanks for re-benchmarking the change! LGTM 👍 

> However, normal numerical parsing was as fast as the original

I think it’s reasonable that the result with a VCF not containing NaN or Inf was on par with before, since the change I suggested is basically only about reducing the overhead of the vector allocation that happens when the submatch occurs, i.e. when the parser encounters a NaN or Inf.

PR #276, opened 2023-06-01
Support explicit first phasing indicator for VCF genotypes

athos:
The current latest VCF spec ([v4.4](https://github.com/samtools/hts-specs/blob/master/VCFv4.4.pdf)) allows explicitly specifying the first phasing indicator of a genotype (like `/0/1` or `|0/1/2`), and defines the semantics for when it's omitted as below:

> The first phasing indicator may be omitted and is implicitly defined as `/` if any phasing indicators are `/` and `|` otherwise.
https://github.com/samtools/hts-specs/blob/0dd3e0d7c6f02b0627a29d6cbab777df07b4daad/VCFv4.4.tex#LL509C5-L509C136

This PR supports explicitly specifying the first phasing indicator to genotypes, and changes the genotype stringifier to emit it if necessary.

### DIFF
--- a/test/cljam/io/vcf/util_test.clj
+++ b/test/cljam/io/vcf/util_test.clj
@@ -81,13 +81,16 @@
     "0/0" [[0 false] [0 false]]
     "0/1" [[0 false] [1 false]]
     "1/1" [[1 false] [1 false]]
+    "/0|1" [[0 false] [1 true]]
+    "|1/0" [[1 true] [0 false]]
     "./." [[nil false] [nil false]]
     "0|0" [[0 true] [0 true]]
     "0|1" [[0 true] [1 true]]
     "1|1" [[1 true] [1 true]]
     ".|." [[nil true] [nil true]]
     "0/1/2" [[0 false] [1 false] [2 false]]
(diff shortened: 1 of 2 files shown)


test/cljam/io/vcf/util_test.clj
```diff
@@ -81,32 +81,37 @@
...
+    "|0/1/2" [[0 true] [1 false] [2 false]]))
 
 (deftest about-stringify-genotype
   (are [?gt ?expected]
        (= (vcf-util/stringify-genotype ?gt) ?expected)
     nil nil
     [[0 true]] "0"
     [[1 true]] "1"
-    [[0 false]] "0"
-    [[1 false]] "1"
```
athos, line 99: I dropped these tests for unphased haplotypes (which should be denoted as "/0" or "/1" using the explicit phasing indicator), as I'm not sure if those are valid notationally and conceptually.

alumi (approved):
Thank you for the PR! I confirmed that the new implementation is conforming to the spec. LGTM 👍

xckitahara (approved):
LGTM 👍

Issue #271, opened 2023-05-17
Features for validating VCF/SAM records


athos:
While we are writing out VCF/SAM records to a file, and if we have some invalid data in them, we often end up seeing a confusing error that's pretty hard to tell what was wrong.

It would be nice to have easy-to-understand error messages in order to make it easier to debug errors, but it's highly likely to degrade the serialization performance. So, how about adding features for validating the input VCF/SAM records to see if they are conformant to the format the cljam writer expects, apart from the existing VCF/SAM file IO APIs.

athos:
For VCF, I think the validation criteria should include (but not be limited to):

- Each required field (`CHROM`, `POS`, `ID`, …) has a valid value
- Each `INFO` field has a valid value in terms of the `##INFO` meta info
- Each genotype field has a valid value in terms of the `##FORMAT` meta info and the `FORMAT` field
- If the header has at least one sample column, it also has the `FORMAT` column

athos:
I think it would be convenient to have something like `validate-variants` in the example below so that it can be easily opted in to validate variants when writing out them to a file:

```clojure
(with-open [r (vcf/reader "in.vcf.gz")
            w (vcf/writer "out.vcf.gz" (vcf/meta-info r) (vcf/header r))]
  (->> (vcf/reader r)
       (map process-variant)
       (validator/validate-variants (vcf/meta-info r) (vcf/header))
       (vcf/write-variants w)))
```

Or something like this (but I don't think the validator implementation would be so complicated that we have to manage the internal state as a separate validator instance like this):

```clojure
(with-open [r (vcf/reader "in.vcf.gz")
            w (vcf/writer "out.vcf.gz" (vcf/meta-info r) (vcf/header r))]
  (let [validator (validator/make-validator (vcf/meta-info r) (vcf/header r))]
    (->> (vcf/reader r)
         (map process-variant)
         (validator/validate-variants validator)
         (vcf/write-variants w))))
```

PR #277, opened 2023-06-12
Add validator for writing vcf.

niyarin:
I add VCF4.4 validator for writing.
see also #271

### DIFF
--- a/src/cljam/io/vcf/util/validator.clj
+++ b/src/cljam/io/vcf/util/validator.clj
@@ -1,194 +1,341 @@
 (ns cljam.io.vcf.util.validator
-  (:require [cljam.io.vcf.util :as vcf-util]
-            [clojure.string :as cstr]))
+  (:require [cljam.io.vcf.util :as vcf.util]))
 
-(defn- valid-ref? [s]
-  (and (string? s)
-       (seq s)
-       (every? #{\A \T \G \C \N
-                 \a \t \g \c \n} s)))
+(defrecord VCFValidator [ploidy file-type defs samples])
 
-(defn- valid-qual? [v] (or (nil? v)
-                           (float? v)
-                           (integer? v)))
+(defn- error [path msg & args]
+  {:errors {path [(apply format msg args)]}})
 
-(defn- valid-filter? [filt]
-  (or (nil? filt)
-      (= filt :PASS)
-      (and (sequential? filt)
-           (every? keyword? filt))))
+(defn- warn [path msg & args]
+  {:warnings {path [(apply format msg args)]}})
 
-(defn- validate-alleles [alt vref]
-  (when (sequential? alt)
-    (keep-indexed
-     #(cond
-        (and (not (string? %2)) (not (nil? %2)))
-        (format "An allele must be string but got other type at index %d."
-                %1)
+(defn- merge-validation-results
+  ([] nil)
+  ([res] res)
+  ([res1 res2]
+   (letfn [(rec [x y]
+             (cond (nil? x) y
+                   (nil? y) x
+                   (map? x) (merge-with rec x y)
+                   :else (into x y)))]
+     (rec res1 res2)))
+  ([res1 res2 res3 & more]
+   (reduce merge-validation-results res1 (list* res2 res3 more))))
 
-        (= %2 "")
-        (format "An allele cannot be empty but got an empty allele at index %d."
-                %1)
+(defn- prepend-keys [res keys]
+  ;; takes {:errors {[:chr] ["foo"]}}
+  ;; returns {:errors {[key ... :chr] ["foo"]}}
+  (when-not (nil? res)
+    (into {}
+          (map (fn [[kind m]]
+                 [kind
+                  (into {} (map (fn [[path messages]]
+                                  [(into keys path) messages]))
+                        m)]))
+          res)))
 
-        (= :other (:type (vcf-util/inspect-allele vref %2)))
-        (format "Contains bad allele at index %d." %1))
-     alt)))
+(defn- error-on-bcf [validator path & args]
+  (let [f (if (= :bcf (:file-type validator)) error warn)]
+    (apply f path args)))
 
-(defn- check-base-records [vcf-or-bcf variant meta-included-contig?
-                           meta-included-filter?  meta-included-format?]
-  (reduce-kv
-   (fn [m k vs]
-     (reduce
-      (fn [m [pred msg]]
-        (cond-> m (not (pred (k variant))) (update k (fnil conj []) msg)))
-      m
-      vs))
-   nil
-   {:chr
-    [[(complement (partial some #{\< \> \, \space \tab}))
-      (format "Must not contain whitespaces commas or angle brackets, but got %s."
-              (str (vec (distinct (keep #{\< \> \, \space \tab} (:chr variant))))))]
-     [#(or (= vcf-or-bcf :vcf)
-           (meta-included-contig? %))
-      (format (str "CHROM %s must be declared in a "
-                   "##contig field in the header.")
-              (:chr variant))]],
-    :pos [[integer? "Position must be Integer."]],
-    :ref [[valid-ref? "Must consist of ACGTNacgtn."]]
-    :alt [[(every-pred seq sequential?) "Must be a sequence."]
-          [#(not (seq (validate-alleles  % (:ref variant))))
-           (cstr/join ", " (validate-alleles (:alt variant) (:ref variant)))]]
-    :qual [[valid-qual? "Qual must be float."]]
-    :filter [[valid-filter? "Invalid filter."]
-             [#(or (= vcf-or-bcf :vcf)
-                   (every? meta-included-filter? %))
-              (and (sequential? (:filter variant))
-                   (format "Filter identifiers %s must in meta-info."
-                           (vec (remove meta-included-filter?
-                                        (:filter variant)))))]]
-    :format [[meta-included-format? "Must be contain meta"]]}))
+(defn- prep-contig-defs [_validator contigs]
+  (into {} (map (juxt :id identity)) contigs))
 
-(defn- check-entry-type [entry type-str]
-  ((case type-str
-     "Integer" integer?
-     "Float" #(or (float? %) (integer? %))
-     "Character" char?
-     "String" string?
-     "Flag" (constantly true)
-     (constantly false))
-   entry))
+(def ^:private ^:const contig-pattern
+  #"[0-9A-Za-z!#$%&+./:;?@^_|~-][0-9A-Za-z!#$%&*+./:;=?@^_|~-]*")
 
-(defn- check-entry [entry {:keys [type number]} ^long alt-num]
-  (let [entries (if (and (= number 1) (not (sequential? entry)))
-                  [entry] entry)
-        type-check (and entry
-                        (not (every? #(check-entry-type
-                                       % type) entries)))
-        number' (case number
-                  "A" alt-num
-                  "R" (inc alt-num)
-                  "G" (count (vcf-util/genotype-seq 2 alt-num))
-                  number)
-        number-check (not (or (nil? entry)
-                              (= (count entries) number')))]
-    (when (or type-check number-check)
-      (cond-> nil
-        type-check
-        (conj
-         (format
-          "Not match type declaration. Requires %s, but got %s."
-          type (str entries)))
-        number-check
-        (conj
-         (format "Invalid number of elements. Requires %s, but got %d."
-                 (str number) (count entries)))))))
+(defn- validate-chrom [validator {:keys [chr]}]
+  (if (and (string? chr) (re-matches contig-pattern chr))
+    (when-not (get-in validator [:defs :contig chr])
+      (error-on-bcf validator [:chr] "Contig %s not defined in meta info" chr))
+    (error [:chr]
+           (str ":chr must be a non-empty string that consists of characters "
+                "other than whitespace, commas or angle brackets, but got "
+                (pr-str chr)))))
 
-(defn- check-each-samples [variant samples mformat]
-  (reduce
-   (fn [res [ks v]] (assoc-in res ks v))
-   nil
-   (for [sample samples
-         [id entry] (get variant sample)
-         :let [fmt (get mformat id)
-               res (if fmt
-                     (check-entry entry fmt
-                                  (count (:alt variant)))
-                     [(format "Key %s not in meta." id)])]
-         :when res]
-     [[sample id] res])))
+(defn- validate-pos [_validator {:keys [pos]}]
+  (when-not (integer? pos)
+    (error [:pos] (str ":pos must be integer, but got " (pr-str pos)))))
 
-(defn- make-contig-validator [contig]
-  (let [contigs (into #{} (map :id) contig)]
-    #(or (nil? %) (contigs %))))
+(defn- validate-ref [_validator {r :ref}]
+  (when-not (and (string? r) (re-matches #"(?i)[ATGCN]+" r))
+    (error [:ref]
+           (str ":ref must be a non-empty string consisting of A/C/G/T/N "
+                "(case insensitive), but got " (pr-str r)))))
 
-(defn- make-header-contained-filter-validator [filter]
-  (let [filter-ids (into #{} (map (comp keyword :id)) filter)]
-    #(or (nil? %) (= % :PASS) (filter-ids %))))
+(defn- validate-alt [_validator {:keys [alt] :as variant}]
+  (if (or (nil? alt) (sequential? alt))
+    (some->> alt
+             (keep-indexed
+              (fn [i allele]
+                (cond
+                  (and (not (nil? allele))
+                       (or (not (string? allele)) (= allele "")))
+                  {[:alt i]
+                   [(str "Every allele in :alt must be a non-empty string or nil, "
+                         "but got " (pr-str allele))]}
 
-(defn- make-format-validator [format]
-  #(or (nil? %) ((set (map (comp keyword :id) format)) %)))
+                  (and (string? (:ref variant))
+                       (= :other (:type (vcf.util/inspect-allele (:ref variant) allele))))
+                  {[:alt i] [(str "Unexpected allele found: " allele)]})))
+             seq
+             (apply merge)
+             (array-map :errors))
+    (error [:alt]
+           (str ":alt must be a sequence of strings or nil, but got " (pr-str alt)))))
 
-(defn- make-info-validator [meta-info]
-  (let [id->info (into {} (map (juxt (comp keyword :id) identity)) meta-info)]
-    (fn [info alt-num]
-      (reduce-kv
-       (fn [validated id entry]
-         (if-let [info (get id->info id)]
-           (if-let [res (check-entry entry info alt-num)]
-             (assoc validated id res)
-             validated)
-           (assoc validated
-                  id
-                  [(format (str "Key %s is not contained in ##info fields in "
-                                "the header.") id)])))
-       nil
-       info))))
+(defn- validate-qual [_validator {:keys [qual]}]
+  (when-not (or (nil? qual) (number? qual))
+    (error [:qual]
+           (str ":qual must be a number or nil, but got " (pr-str qual)))))
+
+(defn- prep-filter-defs [_validator filters]
+  (into {} (map (juxt (comp keyword :id) identity)) filters))
+
+(defn- validate-filter [validator {filt :filter}]
+  (if (sequential? filt)
+    (->> (for [[i filt] (map-indexed vector filt)]
+           (if (and (keyword? filt)
+                    (re-matches #"[^ \t;]+" (name filt)))
+             (when-not (get-in validator [:defs :filter filt])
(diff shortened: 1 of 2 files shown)


src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,124 @@
...
+    (not (valid-ref? (:ref variant))) (assoc :ref "not include invalid char.")
+    (and (seq (:alt variant))
+         (not (sequential? (:alt variant)))) (assoc :alt "must be sequence.")
+
+    (and (sequential? (:alt variant))
+         (not (every? #(vcf-util/inspect-allele % (:ref variant))
+                      (:alt variant))))
+    (assoc :alt "Invalid allele.")
+
+    (not (valid-qual? (:qual variant))) (assoc :qual "qual must be float.")
```
alumi, line 35:
```suggestion
    (not (valid-qual? (:qual variant)))
    (assoc :qual "Must be a floating point number.")

```

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,124 @@
...
+(defn- valid-filter? [filt]
+  (or (nil? filt)
+      (= filt :PASS)
+      (and (sequential? filt)
+           (every? keyword? filt))))
+
+(defn- check-base-records [variant meta-included-contig? meta-included-format?]
+  (cond-> nil
+    (some #{\< \> \, \  \tab} (:chr variant))
+    (assoc :chr "must not contain whitespaces commas or angle brackets.")
```
alumi, line 22: If multiple messages for the same key can be proposed at the same time, it should be a sequence.
```suggestion
    (update :chr (fnil conj []) "must not contain whitespaces commas or angle brackets.")
```

test/cljam/io/vcf/util/validator_test.clj
```diff
@@ -0,0 +1,132 @@
...
+  (testing "format check"
+    (is (not
+         (validator/invalid-variant?
+          {:chr "chr10" :pos 10
+           :FORMAT ["A"]}
+          {:contig [{:id "chr10"}]}
+          base-header)))))
+
+(deftest make-validator-test
+  (is (thrown? clojure.lang.ExceptionInfo
```
alumi, line 126: Check the expected ex-data explicitly. This helps users finding out the usage of the validator a lot.

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,124 @@
...
+     (constantly false))
+   entry))
+
+(defn- check-entry [entry {:keys [type number]}]
+  (let [entries (if (and (= number 1) (not (sequential? entry)))
+                  [entry] entry)
+        type-check (and entry
+                        (not (every? #(check-entry-type
+                                       % type) entries)))
+        number-check (not (or (= number "A")
```
alumi, line 56: Also validate the number of entries for `A`, `R`, and `G`.
- A: `(count :alt)`
- R: `(inc (count :alt))`
- G: `(count (cljam.io.vcf.util/genotype-seq 2 (count :alt)))` or `(inc (cljam.io.vcf.util/genotype-index [(count :alt) (count :alt)]))`

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,124 @@
...
+
+(defn make-validator [meta-info header]
+  (let [validator (make-validator* meta-info header)]
+    (fn [variants]
+      (map (fn [v]
+             (when-let [info (validator v)]
+               (throw (ex-info "Invalid variant."
+                               (assoc info :variant v))))
+             v)
+           variants))))
```
alumi, line 124:
```suggestion
(defn validate-variants [meta-info header variants]
  (let [validator (make-validator* meta-info header)]
    (map (fn [v]
           (when-let [info (validator v)]
             (throw (ex-info "Invalid variant." (assoc info :variant v))))
           v) variants)))
```

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,149 @@
...
+      (reduce
+       (fn [validated [id entry]]
+         (if-let [info (get id->info id)]
+           (if-let [res (and info
+                             (check-entry entry info alt-num))]
+             (assoc validated id res)
+             validated)
+           (assoc validated
+                  id
+                  (format "Key %s not contain in meta." id))))
```
alumi, line 120:
```suggestion
           (assoc
            validated id
            (format
             "Key %s is not contained in ##info fields in the header." id))))
```

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,149 @@
...
+
+(defn- make-contig-validator [contig]
+  (let [contigs (into #{} (map :id) contig)]
+    #(or (nil? %) (contigs %))))
+
+(defn- make-format-validator [format]
+  #(or (nil? %) ((set (map (comp keyword :id) format)) %)))
+
+(defn- make-info-validator [meta-info]
+  (let [id->info (map (comp keyword :id) meta-info)]
```
alumi, line 109:
```suggestion
  (let [id->info (into {} (map (juxt (comp keyword :id) identity)) meta-info)]
```

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,149 @@
...
+    (when (or type-check number-check)
+      (cond-> nil
+        type-check
+        (conj
+         (format
+          "Not match type declaration. Requires %s , but %s."
+          type (str entries)))
+        number-check
+        (conj
+         (format "Invalid number of elements. Requires %s , but %d."
```
alumi, line 84:
```suggestion
         (format "Invalid number of elements. Requires %s, but got %d."
```

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,149 @@
...
+                                             alt-num))
+                      :else number)
+        number-check (not (or (nil? entry)
+                              (= (count entries) number')))]
+    (when (or type-check number-check)
+      (cond-> nil
+        type-check
+        (conj
+         (format
+          "Not match type declaration. Requires %s , but %s."
```
alumi, line 80:
```suggestion
          "Not match type declaration. Requires %s, but got %s."
```

alumi (commented):
Thank you for updating. It's getting much better! 👍

alumi (commented):
Thank you for your continuous efforts in improving this! 🙏 I added some more comments.
Also, some of my previous comments were overlapped by one of them and may have been missed, so please take them into consideration again.
- https://github.com/chrovis/cljam/pull/277#discussion_r1262503655
- https://github.com/chrovis/cljam/pull/277#discussion_r1263674474
- https://github.com/chrovis/cljam/pull/277#discussion_r1264677680
- https://github.com/chrovis/cljam/pull/277#discussion_r1264677766
- https://github.com/chrovis/cljam/pull/277#discussion_r1264678983

athos (commented):
Thanks for drafting this PR and sorry for the terribly late reaction to it!

Personally, I'm thinking that the VCF validator namespace should provide functions like the following as a public API, from the aspects of consistency and extensibility:

- `(make-validator <meta-info> <header> <opts>)`
- `(validate-variant <validator> <variant>)`
- `(validate-variants <validator> <variants>)`

## Consistency

`invalid-variant?` (btw I'd prefer `valid-variant?`, but anyway) does more than returning a boolean value, so it shouldn’t be defined as a predicate. Rather, I think It would be nice to align its semantics with `validate-variants`’s and provide a functionality that validates a single variant and reports errors if any.

## Extensibility

If we add more validation options in the future, it would be tedious to pass those options to every call to `validate-variant(s)` (especially when they are used in more than one place). So, I think the validation options should be passed to `make-validator`, and `validate-variant(s)` should use the options specified for the validator.

Also, `validate-variants` could be extended to return a transducer (from the arity without variants), but it already has an optional argument `vcf-or-bcf`, which makes it a little bit harder to add another arity for the transducer. In this respect, I think `vcf-or-bcf` should be passed to `make-validator` as another option, such as `:file-type`.

We are highly likely to want to add more options for VCF validation in the future, and a validator will be a good place to put together various information including options.

athos:
JFYI : If we really need to have a  function that returns a map containing error information, I’d prefer adding another "verb" for the different semantics. For example:

- `(validate-variant <validator> <variant>)`: Returns a map that contains error information if any. Otherwise returns nil.
- `(check-variant <validator> <variant>)`: Throws an error if the variant is invalid.

niyarin:
I renamed the api and change `make-validator` to  public.

athos:
Thanks for the updates!

And sorry that I carelessly suggested using the word "check" without paying attention to the code you already wrote, but currently the word "check" is used interchangeably with the word "validate" (like `check-base-records`, `check-entry-type` etc.). We should draw a clear line between how we use those two words.

athos (changes requested):

athos:
Just rewrote the validator implementation.

It should provide basic validation features, but I think it's a bit vulnerable to broken definitions of Info and Genotype fields themselves, so we might want to work on that area as a future work.

athos (approved):
Approved this, just for form's sake.

alumi (approved):
@athos Thank you for organizing the API! 🎉 
It appears to be consistently structured and more user-friendly than before.
The internal implementation also has a greater sense of uniformity and is well-structured.
I agree with this approach, so I approve the PR

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,338 @@
...
+             (cond (nil? x) y
+                   (nil? y) x
+                   (map? x) (merge-with rec x y)
+                   :else (into x y)))]
+     (rec res1 res2)))
+  ([res1 res2 res3 & more]
+   (reduce (fn [res res']
+             (merge-validation-results res res'))
+           res1
+           (list* res2 res3 more))))
```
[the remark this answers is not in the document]
athos: Fixed by ff209fc04114cf70b3bd1defe132d8c463eef506 💪

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,335 @@
...
+             (error [:FORMAT i]
+                    (str "Every genotype key in :FORMAT must be keyword, "
+                         "but got " (pr-str key)))))
+         (apply merge-validation-results))
+    (error [:FORMAT]
+           (str ":FORMAT must be a sequence of keywords or nil, but got "
+                (pr-str fmt)))))
+
+(defn- validate-samples [validator variant]
+  ;; TODO: The presence of the GT field should be checked if :FORMAT is present
```
matsutomo81, line 220: [VCFv4.4 Specification](https://github.com/samtools/hts-specs/blob/9ea3a9d1fe1f18579f9a429ef224a8d05dbc48f1/VCFv4.4.tex#L443) says:
> The first key must always be the genotype (GT) if it is present.

There seems to be a discrepancy between this `TODO:` comment and the specification. Is this comment written with that in mind?

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,335 @@
...
+
+(defn- error-on-bcf [validator path & args]
+  (let [f (if (= :bcf (:file-type validator)) error warn)]
+    (apply f path args)))
+
+(defn- prep-contig-defs [_validator contigs]
+  (into {} (map (juxt :id identity)) contigs))
+
+(defn- validate-contig [validator {:keys [chr]}]
+  (if (and (string? chr) (re-matches #"[^<>, \t]+" chr))
```
matsutomo81, line 45: According to [VCFv4.4 Specification](https://github.com/samtools/hts-specs/blob/9ea3a9d1fe1f18579f9a429ef224a8d05dbc48f1/VCFv4.4.tex#L335), I thought it was permissible to use `<`, `>`.

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,335 @@
...
+                (cond
+                  (and (not (nil? allele))
+                       (or (not (string? allele)) (= allele "")))
+                  {[:alt i]
+                   [(str "Every allele in :alt must be a non-empty string or nil, "
+                         "but got " (pr-str allele))]}
+
+                  (and (string? (:ref variant))
+                       (= :other (:type (vcf.util/inspect-allele (:ref variant) allele))))
+                  {[:alt i] [(str "Unexpected allele found: " allele)]})))
```
matsutomo81, line 77: If the [IUPAC ambiguity codes](https://github.com/samtools/hts-specs/blob/9ea3a9d1fe1f18579f9a429ef224a8d05dbc48f1/VCFv4.4.tex#L229-L234) are used, it will result in an error. Is there no current need to consider this?

matsutomo81:
Thank you for creating this PR and I apologize for the significant delay in responding to your review request 🙏 
I haven't been able to review all of the code yet, but I commented on a few points I've noticed so far.

src/cljam/io/vcf/util/validator.clj
```diff
@@ -0,0 +1,335 @@
...
+                                [gt-key f])))
+                   (:FORMAT variant))]
+      (->> (for [sample-name (:samples validator)
+                 [gt-key f] fs]
+             (f sample-name gt-key variant))
+           (apply merge-validation-results)))))
+
+(defn- validate-data-record [validator variant]
+  (if (map? variant)
+    (let [f (juxt validate-contig
```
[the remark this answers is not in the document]
athos: Fixed by adcfe1a 💪

src/cljam/io/vcf/util/validator.clj
athos: I was not really sure what you actually meant by "discrepancy", but anyway I just added some code to check the presence of `:GT` as the first genotype field, rather than leaving it to be a TODO.

src/cljam/io/vcf/util/validator.clj
athos: I'm not sure to what extent cljam supports IUPAC ambiguity codes or symbolic alleles in general, but I don’t think it would report any error in this particular case.

```clojure
(vcf.util/inspect-allele "T" "<R>")
;=> {:type :id, :id "R"}
```

Do you have any concrete example that produces an error?

test/cljam/io/vcf/util/validator_test.clj
```diff
@@ -0,0 +1,470 @@
...
+          validator (assoc validator :defs {:contig prepped})]
+      (are [variant expected]
+           (= expected (#'v/validate-contig validator variant))
+        {:chr "1"} nil
+
+        {:chr 1}
+        {:errors
+         {[:chr]
+          [(str ":chr must be a non-empty string that consists of characters "
+                "other than whitespace, commas or angle brackets, but got 1")]}}
```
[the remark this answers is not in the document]
athos: When a string value is printed in that error message, it will be enclosed in double quotes, by which I think users can tell if the value is a string:

```clojure
(#'validator/validate-chrom nil {:chr " 1 "})
;=> {:errors
;    {[:chr]
;     [":chr must be a non-empty string that consists of characters other than whitespace, commas or angle brackets, but got \" 1 \""]}}
```

Of course, we can add the type of the value to the error message, like `"... , but got 1 (java.lang.Long)"`. But, there are many other similar error messages, and so I think it’s OK to leave them as-is for another PR until we really find them problematic.

src/cljam/io/vcf/util/validator.clj
matsutomo81: Thanks for adding the new commit 👍  I have confirmed that.
What I'm asserting is that the VCF specification permits the absence of the GT field.
Is there a practical reason why the GT field is considered essential?

athos: > What I'm asserting is that the VCF specification permits the absence of the GT field.

Ah, right 😂 In theory, `:FORMAT` is allowed to be empty, so technically my comment wasn't accurate.

> Is there a practical reason why the GT field is considered essential?

I'm not the spec author, so I don't know what the real reason is, but I guess one of the reasons is that if you don't know the genotypes, you can’t tell how many values are necessary for the fields of `Number G` (which require as many values as the genotypes).

src/cljam/io/vcf/util/validator.clj
athos: Thanks for pointing it out! Honestly, I didn't have much time when I was writing the code, so I just borrowed that condition from the original code.

But, after reconsulting the spec, I found the following descriptions:

> It is recommended for VCF, and required for BCF, that the header includes tags describing the contigs referred to in the file.
> https://github.com/samtools/hts-specs/blob/9ea3a9d1fe1f18579f9a429ef224a8d05dbc48f1/VCFv4.4.tex#L246

> Contig names follow the same rules as the SAM format’s reference sequence names: they may contain any printable ASCII characters in the range [!-\~] apart from ‘\ , "‘’ () [] {} <>’ and may not start with ‘\*’ or ‘=’. Thus they match the following regular expression:
>   [0-9A-Za-z!#$%&+./:;?@^\_|\~-][0-9A-Za-z!#$%&\*+./:;=?@^\_|\~-]\*
> https://github.com/samtools/hts-specs/blob/9ea3a9d1fe1f18579f9a429ef224a8d05dbc48f1/VCFv4.4.tex#L266-L271

From these descriptions alone, it’s not clear whether CHROM should follow the naming rule of contigs. However, you can also find the following statement in the spec:

> The CHROM field in BCF2 is encoded as an integer offset into the list of ##contig field headers in the VCF header.
> https://github.com/samtools/hts-specs/blob/9ea3a9d1fe1f18579f9a429ef224a8d05dbc48f1/VCFv4.4.tex#L1975

With this in mind, at least in BCF, CHROM must be a contig defined in the header, which means CHROMs need to obey contig’s naming rule. Ideally, I think it’s the case with VCF as well.

So, to wrap up, I think the regex here should be:

```
[0-9A-Za-z!#$%&+./:;?@^_|~-][0-9A-Za-z!#$%&*+./:;=?@^_|~-]*
```

If you agree with this interpretation, I'll fix the code as such.

src/cljam/io/vcf/util/validator.clj
matsutomo81: I think that according to the specification, it should be allowed to write the IUPAC code defined in `##ALT` in the `ALT` field.
For example:
```
##ALT=<ID=R,Description="IUPAC code R = A/G">
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO
chr1      1     .       C       R       .       PASS    .
```
In other words, I'm concerned that there may be situations where an alphabet other than `ACGTN` should be accepted. For instance, when using the `vcf.util/inspect-allele [ref alt]` function, as demonstrated in `(vcf.util/inspect-allele "T" "R")`, it might need to handle such alphabets in the alt parameter.
However, looking at the [VCFv4.4](https://github.com/samtools/hts-specs/blob/9ea3a9d1fe1f18579f9a429ef224a8d05dbc48f1/VCFv4.4.tex#L2620), I'm wondering if the use of IUPAC ambiguity codes is not prohibited, but not recommended :
> IUPAC ambiguity codes should be converted to a concrete base.

athos: From my understanding, IUPAC ambiguity codes are merely an example use of symbolic alleles.

The spec describes `ALT` as follows, which indicates IDs of symbolic alleles are intended to be used being surrounded in brackets, not to be used as bare IDs:

> Each allele in this list must be one of:
> - a non-empty String of bases (A,C,G,T,N; case insensitive);
> - the `*` symbol (allele missing due to overlapping deletion);
> - the MISSING value `.` (no variant);
> - an angle-bracketed ID String (`<ID>`);
> - the unspecified allele `<*>` as described in Section 5.5; or
> - a breakend replacement string as described in Section 5.4
>
> https://github.com/samtools/hts-specs/blob/9ea3a9d1fe1f18579f9a429ef224a8d05dbc48f1/VCFv4.4.tex#L361

Also, the description below is for `REF`, not for `ALT`, and considering the other context, I think it's about how to resolve symbolic allele abuse in `REF`.

> IUPAC ambiguity codes should be converted to a concrete base.

matsutomo81: Your explanation helped me to understand correctly. Thank you for your explanation 👍

src/cljam/io/vcf/util/validator.clj
matsutomo81: Thank you for your suggestion. 
I think it's reasonable, so I agree to change the code accordingly 👍

src/cljam/io/vcf/util/validator.clj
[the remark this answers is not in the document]
matsutomo81: Shouldn't this test name also require a similar renaming?
https://github.com/chrovis/cljam/blob/adcfe1af7b7c70f878b75ef3a838a07fb28b5ec9/test/cljam/io/vcf/util/validator_test.clj#L34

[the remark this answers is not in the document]
athos: Oh, good catch! Fixed by 28b7303 👍

matsutomo81 (approved):
I’ve completed the code review. 
LGTM! 👍

athos:
Thank you all for reviewing! I just squashed the commits.

PR #279, opened 2023-06-19
Fix/additional coding style

niyarin:
This PR adds clj-kondo's default off linter to clj-lint-action and fixes the warned code.
#273

PR #281, opened 2023-07-06
Fix/shadowing var

niyarin:
This PR fix shadowing variables.
I've fixed what's in the comment (https://github.com/chrovis/cljam/pull/279#pullrequestreview-1489657143 ) here in this PR.
- Variable `meta-info` of `cljam.io.bcf.reader` seemed difficult, so I removed it from the check target.
- Variable `version` of `cljam.io.bcf.reader` could be renamed, but it is also excluded from inspection because the code becomes complicated.

### DIFF
--- a/src/cljam/algo/convert.clj
+++ b/src/cljam/algo/convert.clj
@@ -29,11 +29,11 @@
   (let [refs (refs/make-refs hdr)
         n-threads (get-exec-n-threads)]
     (doseq [blocks (cp/pmap (if (= n-threads 1) :serial (dec n-threads))
-                            (fn [chunk]
+                            (fn [chunk']
                               (mapv #(let [bb (ByteBuffer/allocate (encoder/get-block-size %))]
                                        (encoder/encode-alignment bb % refs)
                                        {:data (.array bb)})
-                                    chunk))
+                                    chunk'))
                             (partition-all num-block (sam/read-alignments rdr {})))]
       (sam/write-blocks wtr blocks))))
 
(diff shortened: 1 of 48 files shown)


xckitahara (approved):
LGTM 👍

alumi (commented):
Thank you for working on this, @niyarin! 👍 
The changes look good for `src`, but `clj-kondo` reports several shadowed vars in `test` directory. Would you fix the local binding names in `tests`, too?

src/cljam/algo/convert.clj
```diff
@@ -13,7 +13,8 @@
             [cljam.io.util :as io-util]
             [com.climate.claypoole :as cp])
   (:import [java.nio ByteBuffer]
-           [cljam.io.fastq FASTQRead]))
+           [cljam.io.fastq FASTQRead])
+  (:refer-clojure :exclude [chunk seq]))
```
alumi, line 17: `chunk` in this ns can be renamed to `chunk'` or something.

https://github.com/chrovis/cljam/blob/eb2e396c84ef45813a5d3806df07dbe0a27c73b5/src/cljam/algo/convert.clj#L33-L37

The same for `seq`.
https://github.com/chrovis/cljam/blob/eb2e396c84ef45813a5d3806df07dbe0a27c73b5/src/cljam/algo/convert.clj#L120

```clj
  [{:keys [flag qual qname] ^String seq' :seq}]
```

In general, I prefer renaming symbols over `:refer-clojure :exclude` because the latter requires programmers to memorize the global state in each ns while reading/modifying codes.

src/cljam/io/bcf/reader.clj
```diff
@@ -13,7 +13,8 @@
   (:import [java.io Closeable IOException FileNotFoundException]
            [java.net URL]
            [java.nio Buffer ByteBuffer]
-           [bgzf4j BGZFInputStream]))
+           [bgzf4j BGZFInputStream])
+  (:refer-clojure :exclude [meta ref]))
```
alumi, line 17: - `meta` at L92 and L255 should be replaced with `meta-info`

alumi (commented):
Would you consider whether there are any other files where we can avoid using the `:exclude` directive by simply changing variable names? 🙏 
I have left comments on some files, so please use them as a reference for your revision strategy.

alumi (commented):
@niyarin Thank you for addressing the previous points!
I apologize for the request again, but could you please consider if it's possible to further change the names of local bindings to minimize the use of `:exclude` directives?
I've added comments on three points as examples.
Please take a look at other codes based on the comments' guidance.
I want to emphasize again that `:exclude` can be used in situations like when it's crucial for the names to match exactly as specified in the documentation.

niyarin:
I fixed shadowing-var in vcf-validator that recently entered master.

alumi (approved):
Thank you, @niyarin! LGTM 👍 
Would you squash your commits before merging?

PR #286, opened 2023-10-19
Use aset instead of aset-* for performance

athos:
`aset-*` (`aset-int`, `aset-long`, etc.) is a set of reflective array operations that are tens of times slower than inline `aset` calls and can have a significant performance impact:

```clojure
user=> (let [arr (int-array 1000)] (cr/quick-bench (dotimes [i 1000] (aset arr i i))))
Evaluation count : 587442 in 6 samples of 97907 calls.
             Execution time mean : 1.012114 µs
    Execution time std-deviation : 5.606968 ns
   Execution time lower quantile : 1.005990 µs ( 2.5%)
   Execution time upper quantile : 1.020545 µs (97.5%)
                   Overhead used : 6.755950 ns
nil
user=> (let [arr (int-array 1000)] (cr/quick-bench (dotimes [i 1000] (aset-int arr i i))))
Evaluation count : 8418 in 6 samples of 1403 calls.
             Execution time mean : 73.125862 µs
    Execution time std-deviation : 1.971431 µs
   Execution time lower quantile : 71.480682 µs ( 2.5%)
   Execution time upper quantile : 75.451133 µs (97.5%)
                   Overhead used : 6.755950 ns
nil
user=>
```

This PR replaces all the invocations to `aset-*` in the codebase with `aset` for performance improvement.

This change contributes specifically to the performance improvement of `cljam.algo.depth`, which is more than 1.5x faster after the change:

```clojure
;; before change
user=> (time (dorun (with-open [r (sam/reader "sample.bam")] (depth/depth r {:chr "chr1"} {:unchecked? true}))))
"Elapsed time: 41775.877042 msecs"
nil
user=>

;; after change
user=> (time (dorun (with-open [r (sam/reader "sample.bam")] (depth/depth r {:chr "chr1"} {:unchecked? true}))))
"Elapsed time: 24674.470916 msecs"
nil
user=>
```

| before change | after change |
| :---: | :---: |
| flamegraph-before-change | flamegraph-after-change |

### DIFF
--- a/src/cljam/algo/depth.clj
+++ b/src/cljam/algo/depth.clj
@@ -23,7 +23,7 @@
             right (min (long (:end aln)) end)
             left-index (- left beg)]
         (dotimes [i (inc (- right left))]
-          (aset-long pile (+ i left-index) (inc (aget pile (+ i left-index)))))))
+          (aset pile (+ i left-index) (inc (aget pile (+ i left-index)))))))
     (seq pile)))
 
 (defn- lazy-depth*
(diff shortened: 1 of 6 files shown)


alumi (approved):
Thank you for sending the performance improvement PR!
I hadn't noticed at all that `aset-*` wasn't being inlined and was calling the reflection API.
I confirmed the speedup here as well using the `cljam.algo.depth-bench` code.
LGTM 👍

niyarin (approved):
LGTM. Thanks!

NoahTheDuke:
This is being tracked in [CLJ-2840](https://clojure.atlassian.net/browse/CLJ-2840), so hopefully such changes won't be necessary in the future.

athos:
@NoahTheDuke  Thank you for letting us know! We are already watching the ticket and hope that fix will further improve Clojure.

However, for this library, similar fixes might still be necessary in the future since we need to support older versions of Clojure as well.

PR #287, opened 2023-10-26
Misc performance tuning for cljam.algo.depth

athos:
This PR includes the following two small changes to improve the performance of `cljam.algo.depth`:

- 736a9bc: Uses `run!` to avoid unnecessary sequence allocation
- 3b15a95: Reads BAM raw blocks into chunked seq, instead of cons, to improve the performance of later processing

The timing results are as follows:

```clojure
(time
 (with-open [r (sam/reader "DRR002191.chr1.bam")]
   (run! (constantly nil)
         (depth/depth r {:chr "chr1"} {:unchecked? true :n-threads 8}))))
;; before the changes
"Elapsed time: 4760.83475 msecs"
;; with 736a9bc
"Elapsed time: 4247.021042 msecs"
;; with 736a9bc + 3b15a95
"Elapsed time: 3983.6545 msecs"
```

The second change only shows a relatively small performance gain and might not need to be included in this PR.

For reference, here are the flamegraphs resulting from profiling:

<details><summary>Profiling results</summary>

- before the changes
Profiling result at b7f7fa1

- with 736a9bc
Profiling result at 736a9bc

- with 736a9bc + 3b15a95
Profiling result at 3b15a95
</details>

### DIFF
--- a/src/cljam/algo/depth.clj
+++ b/src/cljam/algo/depth.clj
@@ -69,14 +69,15 @@
   (let [beg (int beg)
         end (int end)
         offset (int offset)]
-    (doseq [^SAMRegionBlock aln alns]
-      (let [left (Math/max ^int (.pos aln) beg)
-            right (unchecked-inc-int (.end aln))
-            left-index (unchecked-add-int (unchecked-subtract-int left beg) offset)
-            right-index (unchecked-add-int (unchecked-subtract-int right beg) offset)]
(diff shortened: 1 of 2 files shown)


alumi (approved):
Thank you for the improvement!
As for chunking in 3b15a95, I think this modification is acceptable.
LGTM 👍

matsutomo81 (approved):
Thank you for the improvements and the detailed explanation. 
LGTM 👍

Issue #289, opened 2023-11-14
Replacing LSB protocol


athos:
The LSB protocol (i.e., `LSBReadable`/`LSBWritable`) is widely used within cljam as a foundation for encoding and decoding various binary formats.

However, the LSB protocol has some performance issues:

- In general, each method of Clojure protocols has its own method cache, which keeps the overhead of method invocation relatively small as long as the method is called only on one implementation (in a so-called monomorphic situation)
- However, the LSB protocol has multiple implementations, and decoding a file of a particular binary format occasionally involves method invocations to multiple implementations
- Even worse, decoding in parallel causes method invocations to occur in parallel, which disturbs the method cache for each method invocation and results in much more overhead

In particular, reading a BAM file (including reading the BAM index) involves three implementations of the LSB protocol (namely, `DataInput`/`InputStream`/`ByteBuffer`) and has a potentially large overhead. My rough profiling for `cljam.algo.depth`, where a BAM file is read in parallel, shows that the overhead to do with the method invocations to the LSB protocol sums up around 10%.

The LSB protocol does not inherently require dynamic dispatching and has too much overhead to be one of the underpinning parts of cljam’s low layers. I think we should replace the LSB protocol with something more efficient.

A concrete plan to replace the LSB protocol is like the following:

- [x] #290
- [x] #291
  - The current implementation of `TwoBitWriter` is somewhat tricky in that two LSB protocol implementations are used at the same time
- [x] #293

I’m filing three individual PRs for these.

athos:
Closed as completed

PR #290, opened 2023-11-14
Use ByteBuffer wrapper instead of LSB protocol

athos:
This PR is the first step of #289.

It replaces the use of `ByteBuffer` as an LSB protocol implementation with a new thin wrapper utility for `ByteBuffer`.

The modification policy here is to use the bare `ByteBuffer` methods whenever possible. The new wrapper namespace only collects functions that do what a single method of `ByteBuffer` cannot do. Most of the implementation code comes from [the implementation of the LSB protocol](https://github.com/chrovis/cljam/blob/14e0deb16f4423686d5d68907e040fe9729d262d/src/cljam/io/util/lsb.clj#L36-L82).

PR #291, opened 2023-11-22
Rewrite TwoBitWriter using FileChannel and ByteBuffer

athos:
This PR is the second step of #289.

The context of this change is that the current TwoBitWriter uses two implementations of the LSB protocol (i.e., `DataInput`/`FileChannel`), which makes it cumbersome to replace the LSB protocol. So, before doing that, a simplification is needed for `TwoBitWriter` so that it only relies on `FileChannel` (and `ByteBuffer` for buffering).

After rewriting, It turned out that using `FileChannel` instead of `BufferedOutputStream` reduces the relatively large overhead of taking a lock on each `BufferedOutputStream#write`, which improves the performance by 1.5x compared to the current implementation. To put it and some other performance tuning all together, the total performance gain has reached around 3x:

```clojure
(require '[cljam.io.sequence :as seq])

(with-open [r (seq/reader "hg38.2bit")]
  (let [index (vec (seq/read-indices r))
        seqs (vec (seq/read-all-sequences r {:mask? true}))]
    (with-open [w (seq/writer "out.2bit" {:index index})]
      (time (seq/write-sequences w seqs)))))

;; before change
"Elapsed time: 29161.09775 msecs"

;; after change
"Elapsed time: 9473.633458 msecs"
```

In the course of rewriting, I noticed that the current `TwoBitWriter` does not take into account cases where some hard- and soft-mask regions overlap, which ends up producing a non-identical file as compared bit-by-bit to the original one when you read in a .2bit file as complicated as `hg38.2bit` and then write it back to another .2bit file. This PR includes a fix for this issue as well and updates some test code and resources for the .2bit format to ensure that it works right.

<details><summary>Profiling results</summary>

- before change
profiling result before change

- after change
profiling result after change

</details>


athos:
Now that #290 is merged, I've marked this as ready for review.

niyarin (approved):
LGTM!

alumi (approved):
LGTM 👍 Thank you for the great performance improvement and the bug fix!

PR #309, opened 2024-05-10
Support for CRAM beta coding

athos:
This PR adds support for CRAM beta coding.

For more information on beta coding, see [the CRAM specification](https://github.com/samtools/hts-specs/blob/401e2ec3e8ecf19fa89def74e9c27c1e40df341e/CRAMv3.tex#L2147-L2191).

The core of this PR is the `BitStreamDecoder`, which is a bit stream decoder. The `BitStreamDecoder` manages two states `buffer` (the byte value already read) and `nbits` (indicating how many bits of the `buffer` are still "alive") and prepares for reading the next bit sequence.

The CRAM reader creates a `BitStreamDecoder` based on the core block (the only bit-encoded data block) and shares it between data series that are declared to be beta-encoded (Although the CRAM specification does not explicitly state that a core block may be associated with multiple data series, it seems to be guaranteed by the fact that a test file  (`1101_BETA.cram`) provided by [hts-specs](https://github.com/samtools/hts-specs/tree/master/test/cram/3.0/passed) is encoded as such).

### DIFF
--- a/src/cljam/io/cram/bit_stream.clj
+++ b/src/cljam/io/cram/bit_stream.clj
@@ -1,0 +1,31 @@
+(ns cljam.io.cram.bit-stream
+  (:import [java.nio ByteBuffer]))
+
+(defprotocol IBitStreamDecoder
+  (read-bits [_ m]))
+
+(definline ^:private right-nbits-of [x nbits]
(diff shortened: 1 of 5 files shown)


alumi (approved):
Thank you for adding this feature!
I checked both implementation and tests.
LGTM 👍 👍

PR #311, opened 2024-05-15
Support for reading alignments in specific region without CRAI index file

athos:
This PR allows the CRAM reader to read alignments from a specified region within a CRAM file without requiring an index file.

CRAM readers can instantly check if each container or slice overlaps with the specified region by examining their headers. This makes it much more efficient to read alignments by skipping non-overlapping containers or slices, rather than performing a linear scan of the entire CRAM file.

This PR also adds a new test file `medium_without_index.cram`, which is identical to `medium.cram` except that it does not come with its corresponding index file.


test/cljam/io/cram_test.clj
```diff
@@ -39,11 +39,16 @@
...
-                                        {:reference common/hg19-twobit-file})]
+                                        {:reference common/hg19-twobit-file})
+                  cram-rdr' (cram/reader common/medium-without-index-cram-file
+                                         {:reference common/hg19-twobit-file})]
         (is (cram/indexed? cram-rdr))
-        (are [?region ?count] (= ?count (count (cram/read-alignments cram-rdr ?region)))
+        (is (not (cram/indexed? cram-rdr')))
+        (are [?region ?count] (= ?count
+                                 (count (cram/read-alignments cram-rdr ?region))
+                                 (count (cram/read-alignments cram-rdr' ?region)))
```
alumi, line 51: This test fails in my environment.
Maybe we need to reset the position of the mapped buffer before calling next `read-alignments`

athos: Thank you for pointing that out! Yeah, I'm just looking into it right now. I wasn't aware that I didn't run these tests on my local environment at all 😂

athos: b1b0ea4 should fix the position issue.
I also added 3623b37 to improve test coverage for the CRAM region read.

alumi (approved):
Thank you for fixing so quickly! LGTM 👍

niyarin (approved):
LGTM!

PR #312, opened 2024-06-24
Implement ITF-8/LTF-8 encoder for CRAM writer

athos:
This PR introduces the implementation of the ITF-8/LTF-8 encoder as a step towards adding a CRAM writer.

Unlike the existing ITF-8 decoder, which is based on ByteBuffer, the new ITF-8 encoder is implemented using OutputStream. This makes it easier to use without needing to determine the buffer size beforehand.

While adding tests, I found that some of the existing test cases for the ITF-8 decoder used invalidly encoded ITF-8 byte sequences. This PR includes corrections for these test cases.


alumi (approved):
Thank you for the PR! LGTM 👍 👍

PR #318, opened 2024-07-29
CRAM tag dictionary construction at preprocessing phase

athos:
This PR refactors the CRAM writer to incorporate the tag dictionary construction into the preprocessing phase.

Previously, the CRAM writer had a separate pass to scan all records in the container for constructing the tag dictionary. This was done between the pass to encode CRAM records and the preprocessing pass (which is necessary for calculating various header components). This separate scanning pass made the CRAM writer's structure less clear and also caused inefficiency by requiring an additional scan.

By incorporating this scan into the preprocessing phase, the CRAM writer becomes easier to understand and eliminates the inefficiency of the extra scan.

To this end, the PR introduces a _tag dictionary builder_, which scans container records during preprocessing to construct the tag dictionary. It manages IDs for each combination of tags and assigns an ID to each CRAM record based on the combination of tags it has.

### DIFF
--- a/test/cljam/io/cram/encode/record_test.clj
+++ b/test/cljam/io/cram/encode/record_test.clj
@@ -162,7 +162,7 @@
                      :cigar "1M1I1M1D2M", :rnext "=", :pnext 5, :tlen -15, :seq "GAAAG", :qual "EBBFF"
                      :options [{:RG {:type "Z", :value "rg002"}}
                                {:MD {:type "Z", :value "3^T2"}}
-                               {:NM {:type "c",  :value 2}}]}
+                               {:NM {:type "c", :value 2}}]}
                     {:qname "q005", :flag 73, :rname "ref", :pos 20, :end 24, :mapq 0,
                      :cigar "5M", :rnext "*", :pnext 0, :tlen 0, :seq "CTGTG", :qual "AEEEE"
                      :options []}])
--- a/test/cljam/io/cram/encode/record_test.clj
+++ b/test/cljam/io/cram/encode/record_test.clj
@@ -1,6 +1,7 @@
 (ns cljam.io.cram.encode.record-test
   (:require [cljam.io.cram.data-series :as ds]
             [cljam.io.cram.encode.record :as record]
+            [cljam.io.cram.encode.tag-dict :as tag-dict]
             [cljam.io.cram.seq-resolver.protocol :as resolver]
             [cljam.io.sequence :as cseq]
             [cljam.test-common :as common]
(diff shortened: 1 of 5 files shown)


athos:

test/cljam/io/cram/encode/record_test.clj
```diff
@@ -111,46 +140,37 @@
...
-                               {:NM {:type "c", :value 0}}]
-                     ::record/tags-index 1}
+                               {:NM {:type "c", :value 0}}]}
                     {:qname "q004", :flag 147, :rname "ref", :pos 15, :end 19, :mapq 15,
                      :cigar "1M1I1M1D2M", :rnext "=", :pnext 5, :tlen -15, :seq "GAAAG", :qual "EBBFF"
                      :options [{:RG {:type "Z", :value "rg002"}}
                                {:MD {:type "Z", :value "3^T2"}}
-                               {:NM {:type "c",  :value 2}}]
-                     ::record/tags-index 1}
+                               {:NM {:type "c",  :value 2}}]}
```
ToshimitsuArai, line 165:
```suggestion
                               {:NM {:type "c", :value 2}}]}
```

test/cljam/io/cram/encode/record_test.clj
```diff
@@ -213,7 +233,7 @@
 
       (is (= 1 (count (get ds-res :TL))))
       (is (= 13 (get-in ds-res [:TL 0 :content-id])))
-      (is (= [1 1 1 1 0] (seq (get-in ds-res [:TL 0 :data]))))
+      (is (= [0 0 0 0 1] (seq (get-in ds-res [:TL 0 :data]))))
```
ToshimitsuArai, line 236: Why expected data-series-encodings in :TL tag is reversed? I can't find it reasons.

ToshimitsuArai (changes requested):
Thank you for PR 👍 Please check a little bit changes and comment.

test/cljam/io/cram/encode/record_test.clj
athos: The tag dictionary builder assigns sequential IDs in the order it encounters each unique combination of tags for the first time. In this test case, the combination of tags `RG:Z,MD:Z,NM:c` is assigned ID `0` first, followed by the empty tag combination, which is assigned ID `1`.

On the other hand, before this PR the tag dictionary was crafted by hand (with the intention of following lexicographical order), so the order of entries was different than how it is now. I think this difference is trivial.

ToshimitsuArai: Thank you for the comment. I understand the difference.

ToshimitsuArai (approved):
Thanks for the comment and fixing. LGTM 👍

athos:
Thank you for the review!

PR #319, opened 2024-07-30
Refactor CRAM writer to organize encoding values into context maps

athos:
This PR refactors the CRAM writer implementation to simplify how values are passed between the CRAM writer and the record encoder.

Currently, the implementation involves a complex passing of various values and objects between the CRAM writer and the record encoder, including:

- Immutable data: CRAM header, preservation map, tag dictionary, substitution matrix, and data series/tag encodings
- Stateful objects: tag dictionary builder, data series/tag encoders

The current approach results in a CRAM writer cluttered with implementation details that are not within its concern, making the code difficult to maintain.

In this PR, all the above values and objects passed between the CRAM writer and the record encoder are organized into a context map. The context map has the following two types:

- Container context: Used during preprocessing
- Slice context: Used during record encoding

In the future, additional stateful objects, such as the substitution matrix builder and mate reads handler, will also be managed within context maps.

### DIFF
--- a/src/cljam/io/cram/encode/context.clj
+++ b/src/cljam/io/cram/encode/context.clj
@@ -1,0 +1,54 @@
+(ns cljam.io.cram.encode.context
+  (:require [cljam.io.cram.data-series :as ds]
+            [cljam.io.cram.encode.tag-dict :as tag-dict]))
+
+(defn make-container-context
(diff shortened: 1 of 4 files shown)


athos:

niyarin (approved):
LGTM!

PR #321, opened 2024-08-27
Use XZ for LZMA compression in CRAM

athos:
Currently, CRAM files generated by the CRAM writer containing LZMA-compressed blocks cannot be read correctly by samtools, resulting in an error.

## Repro

```clojure

(with-open [r (cram/reader "test-resources/cram/medium.cram"
                           {:reference "hg19.fa"})
            w (cram/writer "medium_lzma.cram"
                           {:reference "hg19.fa"
                            :tag-compressor-overrides (constantly :lzma)})]
  (cram/write-header w (cram/read-header r))
  (cram/write-alignments w (cram/read-alignments r) (cram/read-header r)))
```

```console
$ samtools view -T hg19.fa medium_lzma.cram
[E::lzma_mem_inflate] LZMA decode failure (error 7)
[E::cram_next_slice] Failure to decode slice
samtools view: error reading file "medium_lzma.cram"
$
```

## Cause

According to the CRAM specification, the LZMA compression method used in CRAM is actually based on the XZ format:

> CRAM uses the xz Stream format to encapsulate this algorithm, as defined in `https://tukaani.org/xz/xz-file-format.txt`.
> https://github.com/samtools/hts-specs/blob/ebebbc8c2910ef2d4c5e7119c6f9ffac3bb6a0cb/CRAMv3.tex#L2503

In fact, htsjdk [uses `XZCompressorOutputStream`/`XZCompressorInputStream`](https://github.com/samtools/htsjdk/blob/204a0dbc132739c54387446793a948904fd30715/src/main/java/htsjdk/samtools/cram/compression/LZMAExternalCompressor.java#L44-L61) from commons-compress when the LZMA method is specified.

On the other hand, the current cljam CRAM writer uses `LZMACompressorOutputStream` for LZMA compression. This mismatch causes the generated CRAM files to be unreadable by samtools or htsjdk.

## Change

The PR replaces the use of `LZMACompressorOutputStream`/`LZMACompressorInputStream` with `XZCompressorOutputStream`/`XZCompressorInputStream` for LZMA compression and decompression.

I have confirmed that with this change, CRAM files generated by the CRAM writer containing LZMA-compressed blocks can be successfully read by samtools without errors.

### DIFF
--- a/src/cljam/io/cram/decode/structure.clj
+++ b/src/cljam/io/cram/decode/structure.clj
@@ -115,7 +115,7 @@
                 compressor (case method
                              1 CompressorStreamFactory/GZIP
                              2 CompressorStreamFactory/BZIP2
-                             3 CompressorStreamFactory/LZMA
+                             3 CompressorStreamFactory/XZ
                              (throw
                               (ex-info (str "compression method " method
                                             " not supported")
--- a/src/cljam/io/cram/encode/compressor.clj
+++ b/src/cljam/io/cram/encode/compressor.clj
@@ -2,7 +2,7 @@
   (:import [java.io ByteArrayOutputStream OutputStream]
            [org.apache.commons.compress.compressors.bzip2 BZip2CompressorOutputStream]
            [org.apache.commons.compress.compressors.gzip GzipCompressorOutputStream]
-           [org.apache.commons.compress.compressors.lzma LZMACompressorOutputStream]
+           [org.apache.commons.compress.compressors.xz XZCompressorOutputStream]
            [org.apache.commons.io.output CountingOutputStream]))
 
 (defprotocol ICompressor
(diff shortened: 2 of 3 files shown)


matsutomo81 (approved):
Thank you for your work on this. 🙏
LGTM 👍

alumi (approved):
Thank you for the quick fix! 👍

Since I was not very familiar with the XZ and LZMA formats, I did some research.
XZ is a format where the data block is enclosed by a header that starts with `fd 37 7a 58 5a 00` and a footer that ends with `59 5a`.
LZMA is a format with a header that begins with `5d` followed by the compression level (`00 00 80 00`).
I confirmed that the CRAM file outputted by the modified code indeed has the XZ structure and can be read by samtools.

athos:
Thank you for reviewing the change and clarifying the difference between those formats! That's helpful as I haven't dug that deep into it.

PR #324, opened 2024-10-09
Support reference embedding for CRAM writer

athos:
This PR adds support for reference embedding during CRAM file writing.

## Specification

The CRAM specification for reference embedding is as outlined in the PR for [reference embedding support in the CRAM reader](https://github.com/chrovis/cljam/pull/307#issue-2258310138).

## Usage

This PR adds a new option `:embed-reference?` to enable reference embedding when writing CRAM files. By default, reference embedding is disabled. To enable it:

```clojure

(with-open [w (cram/writer "path/to/cram/file" { ..., :embed-reference? true, ... })]
   ... )
```

Reference embedding is only applicable when writing a CRAM file with the `SO:coordinate` declaration in the CRAM header. If the CRAM header lacks this declaration, an error will be raised. Like [CRAM index generation](https://github.com/chrovis/cljam/pull/317), this `SO:coordinate` check can be skipped by specifying `:skip-sort-order-check? true`:

```clojure
(with-open [w (cram/writer "path/to/cram/file"
                           { ...
                            :embed-reference true
                            :skip-sort-order-check? true
                            ... })]
  ... )
```

Blocks allocated for reference embedding are compressed using `:gzip` by default. The compression method can be overridden via `:ds-compressor-overrides`, as is done for regular data series, by specifying the pseudo data series `:embedded-ref`:

```clojure
(with-open [w (cram/writer "path/to/cram/file"
                           { ...
                            :embed-reference true
                            :ds-compressor-overrides {:embedded-ref :raw}
                            ... })]
  ... )
```

## Implementation

The CRAM writer reserves content ID `1` for blocks used for reference embedding (referred to as "embedded reference blocks"). When `:embed-reference? true` is specified, the CRAM writer writes the reference sequence covered by each slice into the embedded reference block and sets the `:embedded-reference` field in the slice header to content ID `1`. If `:embed-reference? true` is not specified, the blocks associated with content ID `1` are not used.

When reference embedding is enabled, the `RR` flag in the compression header (indicating whether a reference file is required for reading) is always set to `false`.

Additionally, the CRAM writer ensures that no multiple reference (multi-ref) slices are created when reference embedding is enabled. This is because reference embedding is incompatible with multi-ref slices. Allowing multi-ref slices would negate the benefits of reference embedding by partly requiring a reference file when reading the written CRAM file, which would be inconvenient and confusing.

### DIFF
--- a/src/cljam/io/cram.clj
+++ b/src/cljam/io/cram.clj
@@ -72,13 +72,16 @@
         a tag keyword, returns a keyword or a set of keywords representing
         compression method. It may return another function to add more conditions
         for the tag type and/or block encoding.
-    - create-index?: If true, creates a .crai index file in the course of CRAM
-        file writing.
-    - skip-sort-order-check?: When creating a CRAM index for the CRAM file,
-        the CRAM writer, by default, checks if the header is declared as
-        `SO:coordinate` and raises an error if not.
-        If this option is set to true, the CRAM writer will skip the header check
-        and create an index file regardless of the header declaration."
+    - create-index?: If set to true, the CRAM writer creates a .crai index file
+        in the course of CRAM file writing.
+    - embed-reference?: If set to true, the CRAM writer embeds the reference
+        sequences into the resulting CRAM file, allowing it to be read without
+        needing an external reference file.
+    - skip-sort-order-check?: When creating a CRAM index or embedding reference
+        sequences into a CRAM file, the CRAM writer, by default, checks if
+        the header specifies `SO:coordinate` and raises an error if it does not.
+        If this option is set to true, the CRAM writer will skip this header
+        check and proceed regardless of the header's sort order declaration."
   (^CRAMWriter [f] (writer f {}))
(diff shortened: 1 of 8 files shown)


athos:
Here are some figures for writing out `test-resources/cram/medium.cram` with/without reference embedding:

| | file size | elapsed time |
| :--- | ---: | ---: |
| without reference embedding | 540KB | 7.8s |
| with reference embedding (raw) | 3.0GB | 14s |
| with reference embedding (gzip) | 800MB | 440s |

In the current implementation, the default compression method for embedded reference blocks is `:gzip`. However, `:gzip` may not be the most suitable option, and it might be better to switch to a more efficient method as the default in the future (FYI: htslib [only uses relatively fast compression methods](https://github.com/samtools/htslib/blob/2ff207bd16bb3094843395450aa48936dd774ecb/cram/cram_encode.c#L850-L860) for compressing embedded reference blocks).

alumi (approved):
LGTM 👍👍

niyarin (approved):
LGTM!

PR #325, opened 2024-10-17
SNV frequency-based substitution matrix construction for CRAM writer

athos:
This PR introduces a new feature to the CRAM writer, allowing it to construct a substitution matrix that takes into account the observed frequencies of SNVs.

The change introduces a substitution matrix builder, which works similarly to the existing tag dictionary builder used for constructing the tag dictionary. The substitution matrix builder provides two main functions:

- `assign-code!`: This function is called during preprocessing for each :subst read feature in the records and returns a unique base substitution code (BS code) for every pair of ref and alt bases
- `build-subst-matrix`: After preprocessing, this function is called to generate the substitution matrix based on the SNV frequency information gathered by `assign-code!`

Note that since BS codes cannot be assigned until all records within a container are scanned, `assign-code!` initially returns a unique reference (not the actual BS code) for each SNV. When `build-subst-matrix` is invoked, BS codes are assigned for each SNV, and the references are accordingly updated all at once. This implementation is intended to reduce the overhead of scanning each `:subst` read features again to assign the final BS codes after the records are fully processed.

BS code assignment follows a simple frequency-based method: the most frequently observed alt base for each ref base is assigned BS code 0, the next most frequent is assigned code 1, and so forth.


test/cljam/io/cram/encode/subst_matrix_test.clj
```diff
@@ -0,0 +1,27 @@
...
+  (let [builder (subst-mat/make-subst-matrix-builder)
+        a->t (subst-mat/assign-code! builder (int \A) (int \T))
+        a->c (subst-mat/assign-code! builder (int \A) (int \C))
+        a->g (subst-mat/assign-code! builder (int \A) (int \G))
+        a->t' (subst-mat/assign-code! builder (int \A) (int \T))
+        g->c (subst-mat/assign-code! builder (int \G) (int \C))
+        g->t (subst-mat/assign-code! builder (int \G) (int \T))
+        t->n (subst-mat/assign-code! builder (int \T) (int \N))
+        n->c (subst-mat/assign-code! builder (int \N) (int \C))]
+    (is (= {\A {\C 1, \G 2, \T 0, \N 3}
```
alumi, line 15: I thought it might be a bit tricky how `MutableInt` starts as a counter and then turns into a code table after the `build-subst-matrix` invocation, but it's more efficient than splitting them into separate objects, so I think it's fine.
It would be helpful to better describe the behavior if we add test cases for counters like `(is (= 2 @a->t @a->t')) ...` before calling `build-subst-matrix`.

alumi (approved):
Thank you for implementing the feature! LGTM 👍

test/cljam/io/cram/encode/subst_matrix_test.clj
athos: Thank you for the comment! I totally agree.
Just force-pushed the updated commit which added some tests and comments describing how `MutableInt`s are used and updated.

alumi: Thank you! With the added comments and test cases, I find the behavior much easier to understand now👍👍

PR #328, opened 2024-11-19
CRAM writer option for read name omission

athos:
This PR adds a new option to enable read name omission for the CRAM writer.  
By default, read name omission is disabled. To enable it, specify the option `:omit-read-names? true`:

```clojure

(with-open [w (cram/writer "path/to/cram/file" {..., :omit-read-names? true, ...})]
  ...)
```

### Specification  
According to the CRAM specification, a CRAM encoder may omit encoding read names (`QNAME`), which makes the decoder regenerate them automatically. Details about the CRAM specification have been summarized in a [previous PR](https://github.com/chrovis/cljam/pull/308).

Note that read name omission relies on [non-detached mate record encoding](https://github.com/chrovis/cljam/pull/326). This means that read names can only be omitted for records encoded as non-detached mates. Records encoded as detached, including those for single-end reads, cannot have their read names omitted.  

### Implementation  
When the option `:omit-read-names? true` is specified, the CRAM writer sets the `RN` field in the compression header to `true` and attempts to omit encoding read names for each record.  

Read name omission applies only to non-detached mate records consisting of primary and representative alignments. For other mate records, such as those involving secondary or supplementary alignments, read names will still be written to the `RN` data series, even if `:omit-read-names? true` is specified.  

There is no surefire way to determine whether a set of mate records involves secondary or supplementary alignments. cljam's CRAM writer identifies the presence of secondary alignments using the `TC` tag and supplementary alignments using the `SA` tag (which is the same approach taken by [htslib](https://github.com/samtools/htslib/blob/c384c816a1d94504ab890697dd7c2ab7324589f6/cram/cram_encode.c#L1301-L1319)).

However, it's still the user's responsibility to ensure that these tags are appropriately attached to records so that the writer can correctly identify secondary and supplementary alignments.


athos:

alumi (approved):

niyarin (approved):
LGTM!
Thank you.

PR #329, opened 2024-12-23
Add a function to read mate alignments

alumi:
#### Summary
This PR adds a function `cljam.io.sam/read-mate-alignments` for BAM file which reads R1/R2 counterpart alignments of given alignments.

#### Added functions
This function can be used for collecting alignments around some regions keeping pairs like `samtools view --fetch-pairs`.

```clj
(let [xs (sam/read-alignments reader {:chr "chr1", :start 1, :end 100000})]
  (concat xs (sam/read-mate-alignments reader xs))
```

In the example above, it is necessary to recheck which alignment pairs with which. To make things easier, I added another function `cljam.io.sam/make-pairs` that returns a sequence with paired alignments grouped together.

```clj
(let [xs (sam/read-alignments reader {:chr "chr1", :start 1, :end 100000})]
  (sam/make-pairs reader xs))
```

To achieve these functionalities, I implemented a function `cljam.io.bam-index/get-spans-for-regions` that queries chunks corresponding to multiple regions at once against the BAI index.

#### Implementation details
`cljam.io.sam/read-mate-alignments` for SAM and CRAM is not supported yet.

When searching for mates, a large number of alignment blocks that do not meet the criteria must be discarded. Therefore, the current implementation decodes only the minimal set of fields necessary for condition evaluation and immediately rejects blocks that do not satisfy the criteria.

As a result, a significant part of the execution time is spent decompressing BGZF blocks, so reducing the number of chunks accessed would make a big difference.
However, the current implementation is limited to what can be achieved by combining existing functionalities, leaving room for further optimization in the future.

#### Tests
Since we don't have a good BAM file with paired reads of appropriate size, I added auto-generated test cases for `cljam.io.sam/read-mate-alignments`. It writes a temporary BAM file containing paired alignments on the fly, and then perform reading and checking.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -5,7 +5,7 @@
             :url "https://www.apache.org/licenses/LICENSE-2.0"}
   :dependencies [[org.clojure/core.memoize "1.1.266"]
                  [org.clojure/tools.logging "1.3.0"]
-                 [org.apache.commons/commons-compress "1.26.1"]
+                 [org.apache.commons/commons-compress "1.27.1"]
                  [org.tukaani/xz "1.9"] ; necessary for CRAM LZMA compression
                  [digest "1.4.10"]
                  [bgzf4j "0.1.2"]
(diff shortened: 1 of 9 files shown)


ToshimitsuArai (approved):
Thank you for your PR !
LGTM👍

alumi:
@ToshimitsuArai Thank you for your reviewing! ❤️

PR #331, opened 2025-01-07
Explicitly install lein with setup-clojure for CI

athos:
As mentioned in https://github.com/chrovis/cljam/pull/330#issuecomment-2572767463, Leiningen has been removed from the latest `ubuntu-latest` GitHub Action runner image. Some workflows in cljam rely on the `lein` installed in the `ubuntu-latest` image, which causes them to fail on the CI.

This PR updates those workflows by explicitly installing Leiningen with `DeLaGuardo/setup-clojure`. It also adds `actions/setup-java` to the `coverage` workflow to avoid depending on the `java` installed in the GA runner image, as in the other workflows.


alumi (approved):
LGTM 👍

PR #335, opened 2025-03-11
Change the URLs of the remote test resources

alumi:
I have updated the URLs of the test files due to the replacement of the hosting server.

- 914ad72: The URL update
- 1108d29: Fix kibit warnings
- d9883ef: Fix clj-kondo warnings by new linters of #333

PR #336, opened 2025-03-12
Fix the URLs of the remote test resources again

alumi:
The files of #335 are moved to `/data`.
https://github.com/chrovis/cljam/pull/335#issuecomment-2716620907


alumi:
@r6eve Sorry for the extra work, and thank you for your help! 🙇🙇

Issue #332, opened 2025-01-08
Overflow issues that occur when *unchecked-math* is disabled


athos:
As reported in https://github.com/chrovis/cljam/pull/330#pullrequestreview-2522682459, when running tests with `*unchecked-math*` disabled, you'll see some tests fail due to overflow caused by the `byte` fn:

```console
$ lein update-in : assoc :global-vars '^:replace {*warn-on-reflection* nil, *unchecked-math* nil}' -- test
...
Ran 372 tests containing 3832 assertions.
1 failures, 17 errors.
```

PR #337, opened 2025-03-24
Fix overflow issues occurs when *unchecked-math* is disabled

alumi:
fixes: #332 

This PR fixes issues found in https://github.com/chrovis/cljam/pull/330#pullrequestreview-2522682459

Some functions under `cljam.io.cram` and `cljam.io.util.lsb` used the **checked** narrowing conversions of numerical values (`byte`, `int`, ...) or implicit conversions with typed method callings of Java core library.
These narrowing of numeric types works fine as long as `*unchecked-math*` is set to `true` or only non-negative integers are involved. However, if `*unchecked-math*` is `false` (which is the default) and a negative integer is provided, an exception is thrown.
In this PR, I have explicitly applied unchecked narrowing (`unchecked-byte`, `unchecked-int`, ...) to bypass range checks and ensure the intended output.
I have also fixed the `:gloval-vars` option in the project to reset the compiler options while testing in CI.

I have confirmed that `lein test :all` passes.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -32,7 +32,9 @@
              :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
              :1.9 {:dependencies [[org.clojure/clojure "1.9.0"]]}
              :1.10 {:dependencies [[org.clojure/clojure "1.10.3"]]}
-             :1.11 {:dependencies [[org.clojure/clojure "1.11.4"]]}}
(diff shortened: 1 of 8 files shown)


athos (approved):
Thanks for fixing this (and sorry for embedding lots of subtle potential bugs 😅)! LGTM 👍

alumi:
@athos Thank you for reviewing my PR! 🙏

PR #338, opened 2025-03-25
Fix some linter errors

alumi:
This PR fixes coding styles of test codes and does not introduce any changes of its behavior.

`clj-kondo` reports `test/cljam/io/gff_test.clj:319:70: error: Unresolved symbol: ?str` on the master branch.
I've removed the type hints in 909e6d2cc46fd00dfdd6d917b4a12efd6f9ecaad because adding type hints to the symbols  in `expr` of `clojure.test/are` has no effects on suppressing reflections.
In f13581a936505c98ba2028711e3d871e610057dc, I've also fixed some coding styles warned by `kibit`.

<details>

The strange behavior of the `:unresolved-symbol` error is due to differences in how metadata is handled.
[`clojure.test/are` uses `clojure.template/do-template`](https://github.com/clojure/clojure/blob/fb22fd778a272b034684a4ee94509552b46ee8a9/src/clj/clojure/test.clj#L594) which [replaces symbols using `clojure.walk/postwalk-replace`](https://github.com/clojure/clojure/blob/fb22fd778a272b034684a4ee94509552b46ee8a9/src/clj/clojure/template.clj#L43).
Since a symbol’s metadata does not affect its identity during replacement, this process works as expected.
However, [`clj-kondo` emulates `clojure.template/do-template` with `clojure.walk/postwalk-replace`, but it operates on AST nodes](https://github.com/clj-kondo/clj-kondo/blob/cd5fa583706ba1e2951ca8a8225777e35a4640af/src/clj_kondo/impl/macroexpand.clj#L201), which have `:meta` fields directly attached.
This difference causes replacement failure.
As seen above, since `clojure.test/are` replaces symbols in its `expr`, we need type information on `args` rather than on the symbols in `expr`.

```clojure
user> (set! *warn-on-reflection* true)
user> (def s "A")
user> (are [x] (.length ^String x) s)
Reflection warning, *cider-repl chrovis/cljam:localhost:64423(clj)*:79:16 - reference to field length can't be resolved.
1
user> (are [x] (.length x) ^String s)
1
```

In the case of this PR, adding type hints is unnecessary because `args` are already typed vars.

</details>

### DIFF
--- a/test/cljam/io/gff_test.clj
+++ b/test/cljam/io/gff_test.clj
@@ -316,7 +316,7 @@
 (deftest read-features
   (are [?str ?edn]
        (= ?edn
-          (with-open [bais (ByteArrayInputStream. (.getBytes ^String ?str))
+          (with-open [bais (ByteArrayInputStream. (.getBytes ?str))
                       r (gff/reader bais)]
             (doall (gff/read-features r))))
     simple-gff simple-edn
@@ -364,7 +364,7 @@
   (are [?edn ?str]
       ;; ignore directives and comment lines
        (= (cstr/replace ?str #"(?<=\n)#.*?\n" "")
-          (with-open [bais (ByteArrayInputStream. (.getBytes ^String ?str))
+          (with-open [bais (ByteArrayInputStream. (.getBytes ?str))
                       baos (ByteArrayOutputStream.)]
             (let [v (with-open [r (gff/reader bais)]
                       (gff/version r))]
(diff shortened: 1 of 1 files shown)


niyarin (approved):
Thank you for the fix.
I confirmed that the issue with the linter warning in `test/cljam/io/gff_test.clj`  has been resolved.

alumi:
@niyarin Thank you for reviewing! 🙏

PR #354, opened 2026-08-04
Stop using cache in CI

alumi:
Removed `actions/cache` from `.github/workflows`.


r6eve (approved):
I've confirmed there are no caches now. https://github.com/chrovis/cljam/actions/caches
Thank you for the PR.

alumi:
@r6eve Thank you for your quick review! 🙏

